PR #1, opened 2015-06-02
add video icon to index view gallery for articles with videos

campgurus:
just adding an image file for the video icon and changing the code (including inline styles that we should eventually change) so the user can see which cases have videos attached.


campgurus:
taking this back.  i see a layout issue with this branch now.

PR #48, opened 2015-07-16
adding carrierwave optimization.  set all but one version to quiet lo…

campgurus:
…gs, also edited ahoy initializer back to default code to avoid error


campgurus:
This feature has been integrated.

PR #59, opened 2015-07-23
Jquery.Turbolinks fix

parcheesy:
-added jquery.turbolinks gem to deal with javascript not loading when navigating between pages
-also added it to asset pipeline
-eliminated previous redundant fix for javascipt


parcheesy:
Added Install Readme to help contributors get local environment set up.

campgurus:
I attempted to merge this fix, but was unable to do so.  We unwound some of Mark's changes to search which I had incorrectly pushed to master/production. So now, I think you may need to do the pull request from your branch to the current master which removes Mark's work on search. Mark may know a better way.

PR #96, opened 2015-08-19
Add redis caching and rack attack.

parcheesy:
Added redis caching and rack attack, including rspec tests. The tests won't pass unless ENV["REDISTOGO_URL"] is set to a redis server url in ".env" file for development.

Tested on staging. Before merging to heroku must add on redistogo with:

$ heroku addons:create redistogo

Will work after that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,9 @@ gem 'sdoc', '~> 0.4.0', group: :doc
 # Use Capistrano for deployment
 # gem 'capistrano-rails', group: :development
 
+# Redis for caching 
+gem 'redis-rails'
+
 # Simple Captcha for signing up
 gem 'gotcha'
 
@@ -116,6 +119,12 @@ gem 'gmaps4rails'
 
 # implement elasticsearch
 gem 'searchkick'
+
+# Rack Attack for basic security through throttling and blocking
+gem 'rack-attack'
+
+# Rack Test for testing 
+gem 'rack-test', require: 'rack/test'
 
 # manage nested forms
 gem 'cocoon'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -374,6 +374,8 @@ GEM
     puma (2.11.2)
       rack (>= 1.1, < 2.0)
     rack (1.6.0)
+    rack-attack (4.3.0)
+      rack
     rack-pjax (0.8.0)
       nokogiri (~> 1.5)
       rack (~> 1.1)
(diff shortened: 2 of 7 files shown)


campgurus:
Cool! I just created a branch for a/b testing using the split gem which requires redis so timing is perfect.

trystant:
Overall, thanks for this @parcheesy. I had some questions/feedback.

1. Can the REDISTOGO environment var be set to a local Redis server for dev? I'm assuming that that's the case, but I could be wrong.

2. Doing a quick search on Redis options raised this [page](https://news.ycombinator.com/item?id=4755334) which had this paragraph in particular from andrewguenther:

> I have used RedisToGo, and it was one of the worst products I've ever had the displeasure of dealing with.
> I lost contact with the instance associated with my staging server one morning and filed a support ticket immediately. It was almost 4 days before someone responded, even after daily messages to their support team. After requesting my account information, I was simply sent a link to the "How to Connect to your Redis Instance" help page and the ticket was closed. I reopened the ticket and explained that the connection was not the problem, I was then told that there was no problem and the ticket was again closed. Other than in my initial ticket, I was never again asked to explain my issue, or offered any real support. I determined the service was no longer worth my time and booted up an EC2 instance.
Fast-forward about two months. I now receive an email from RedisToGo: "We are planning some scheduled down time that affects all users on server X because of system issues." The message then proceeded to describe the symptoms I had explained to them months before.
> I sent their support team a very polite email requesting an apology for the way my issue was handled. A day later, I received a half-hearted apology and 1 month credit for some advanced feature they offered. I never cashed it in.
> Never, ever, use RedisToGo if you need a reliable Redis instance.

3. The free plan seems to have no backups. From the [Redis To Go](http://redistogo.com/) page: 

>We also offer a free plan: 5 MB memory, 1 Database, but no backups.

We either need a plan with a backup, or we need to have the redis instance backed up somewhere else. I'd like to see a backup put in place based upon this [Redis](http://redistogo.com/documentation/exporting?language=en) article. 

4. @parcheesy, I see that many of the tests involve setting a limit, sending twice the number of requests as that limit and then checking to see whether the last request results in a 429. What happens to the limit + 1 request in that test? Can we see those results in a spec?

campgurus:
here is a SO discussion: http://stackoverflow.com/questions/29285574/how-to-choose-redis-add-on-on-heroku

parcheesy:
1) Yea, for development on my computer I have it set to a local redis
server. You just have to install redis individually and set the environment
variable to a redis://localhost url based on your specific port/database.
You can also set it to our redis-to-go staging server even for local
development.

2)/3) I agree that in a larger production environment it would probably be
better to use an EC2 instance and run our own redis server. But because we
are only currently using it for caching when it comes to rack attack,
RedisToGo should work fine for us. Even if we lost backups it won't matter
because we are simply using it for throttling ip's and email addresses on a
short time scale.

If we plan on implementing actual caching and maybe even switch to a key
value database down the line it will probably be best for us to just use an
ec2 server. I can set that up whenever we decide on that.

4) Sure I can add another spec that tests the limit + 1 case.

trystant:
1. Cool, thanks!

2/3. It would be cool to run our own, but that raises a bigger issue, namely latency between Heroku and EC2 servers. One of the big challenges is that while redis usually gets installed for one thing, the use cases pile up, and I think it's good to be prepared.

4. Cool, thanks much!

trystant:
I'm going to close this for now.

Issue #183, opened 2015-09-05
Adding Split gem for A/B testing


campgurus:
I provisioned the split gem and added a sample test just to see it work.  The current test is measuring click thru rates for the View Maps button on the root page based on changing the color of the button.  Obviously not a key feature, but at this point, I am just confirming that it works. 

It looks like redis is not set up on either heroku environment, so I am going to do that now.

Branch is split_gem

campgurus:
The split gem is currently working on staging.  However, currently, the results dashboard mounted at /split is currently available to any user.  Instead, I am trying to limit access to admin users only.

To do so, I have attempted to apply the following guide: https://www.codementor.io/tips/8132473431/use-custom-routing-constraints-to-limit-access-to-rails-routes-via-warden

However for some reason, the Constraints module suggested by the turtorial is not registering and the error memorialized in Issue #340. 

The error can be reproduced by loading the branch and typing `rake routes`

campgurus:
This is working on production. Closing.

PR #339, opened 2016-02-16
Adding a simple button color test for split gem to work

campgurus:
Adding split gem
Adding test of the color of the view maps button on the index page
Currently, test resets after completion instead of showing a given user
a consistent view, so we may want to change that. Also, Need to organize future tests with a new ruby object class as outlined here:
https://davidtuite.com/posts/a-b-testing-with-split-in-ruby-on-rails


campgurus:
merging to master to address #183

PR #554, opened 2016-05-17
71 version commit

campgurus:


app/models/article.rb
```diff
@@ -63,6 +65,10 @@ def article_date_cannot_be_in_the_future
     end
   end
 
+  def edit_summary
```
trystant: Why this separate method for the summary? Can you not use the summary method itself?

app/models/article.rb
```diff
@@ -33,6 +34,7 @@ class Article < ActiveRecord::Base
   validates :title, presence: { message: "Please specify a title"}
   validates_associated :subjects
   validates :subjects, presence: { message: 'at least one subject is required'}
+  validates :summary, presence: { message: 'Please use the last field at the bottom of this form to summarize your edits to the article in 140 chars or less.'}
```
trystant: Can someone see a summary for each version as it is submitted, or do they only see the latest? Seems like with summary as a direct field on article, we're limiting it to the latter. Let me know if this is off.

app/models/article.rb
campgurus: this is needed to work with the paper_trail gem.  for some reason it needs to be a model method, not merely a model attribute to be recorded as meta data.  Thus the repetitive method is for this line...

`has_paper_trail :ignore => [:summary], :meta => { :comment  => :edit_summary }`

Issue #578, opened 2016-05-31
Add Edit comment text to follower email


campgurus:
We should add the content of the version commit comment to the emails that get sent to followers when a case is updated.  That will hopefully help give them some guidance as to what has changed in the article.

campgurus:
on production

PR #590, opened 2016-06-05
For #578 adding commit summary to update message

campgurus:
This change will add a little more content to messages to followers.


campgurus:
on production.

PR #602, opened 2016-06-06
145 analytics dash

trystant:


app/controllers/analytics_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def index
+    @last_days = 30
+    @full_width_content = true
+    @visits = Visit.where(started_at: @last_days.days.ago..Time.now)
+    @views = Ahoy::Event.where(name: '$view' )
+  end
+
+private
```
trystant: Why is this called "custom_method"? Seems like a generic name for a specific method.

I also think there's a more elegant way to require a user to be an admin. Here's an excerpt from the `config/routes.rb` file, lines 34 - 39
```
mount Split::Dashboard, at: 'split', :anchor => false, :constraints => lambda { |request|
    request.env['warden'].authenticated? # are we authenticated?
    request.env['warden'].authenticate! # authenticate if not already
    # or even check any other condition
    request.env['warden'].user.admin?
  }
```

spec/models/article_spec.rb
```diff
@@ -137,3 +137,28 @@
     end
   end
 end
+
```
trystant: All these monthly specs need specs to cover months that don't have 30 or 31 days, as well as the leap year.

campgurus: I don't think this is an issue, though I understand your point. For our purposes, it doesn't make sense to compare a month with 30 days to one with 28 or 31. So this is designed to compare successive 30 day periods and is only roughly labeled month-over-month.

app/controllers/analytics_controller.rb
campgurus: there is no particular reason to call it custom_method. I can change the name to something more appropriate.

spec/models/article_spec.rb
trystant: So, do you not plan on comparing Feb and Mar, or just random 30 day periods (which still don't divide 365 easily)?

I think it makes sense to cover this before you end up with bad date logic.

campgurus: Not random. I just want to look at the past 30 days vs the 30 days prior. I want to be able to  judge roughly which way we are trending.  If I instead attempt to compare February to March, the comparison is not meaningful because of the 3-day discrepancy.  If we were talking about financial accounting or something that required reporting on specific months, it would be different, but as it is, I don't think it makes sense to do what you are talking about.   I chose 30 vs 30 over idiosyncratic month v month in this context for the simple reason that we need to make a quick, meaningful comparison to determine which way we are going.

PR #735, opened 2016-07-22
718 simple milestones

campgurus:
Please take a look.  Depending on screen size, the icons in the timeline get covered by the text box.  actually, now that i think of it, we might be able to fix that using z-index attribute in css.

We have also talked about adding other categories of events. We could add an event_type attribute the article_milestone model and use that to automate changing the icon and the color of the circle, but I don't think we need that now.


trystant:
We decided to take milestones out for now, so we can remove this PR, correct?

rlgreen91:
Can you bring me up to speed on what milestones are and why we're (temporarily, I guess) removing them?

trystant:
The milestones are part of a proposed timeline view of each case. As opposed to just posting links, we put the events of the case on a timeline and allow the user to journey through each one.  I think it is still a good goal, but we probably need to revamp how to make that happen.

rlgreen91:
That is a good idea.  I will add that to the roadmap.

rlgreen91:
Given that we are redoing our approach to milestones, I will be closing this PR.

Issue #674, opened 2016-07-03
Create sitemap for Agency model


campgurus:
Agency show pages should be indexed in search engines

trystant:
This has been created and the rake task creates a fresh Agency sitemap.

```
2015-07-07 09:36:35          0 sitemaps/
2017-09-06 11:01:05        245 sitemaps/agencies.xml.gz
2017-09-06 11:01:04      23622 sitemaps/agencies1.xml.gz
2015-09-21 12:40:55         53 sitemaps/google949221dfa5e420ae.html
2017-09-06 11:01:00        240 sitemaps/sitemap.xml.gz
2017-09-06 11:01:00      78594 sitemaps/sitemap1.xml.gz

```

PR #753, opened 2016-08-01
for #674 fixes and additions to sitemap config file

campgurus:
I just rebased this branch.  We now have several hundred agency pages that we could get indexed for search. I think this makes sense to do. I believe I have done this correctly to create a separate sitemap for agencies, but want code review before merging. Previous commit message below:

I didn't want to push this without checking in with @trystant first.
This branch contains the following changes to the sitemap config:

1. Added a line to create_index within the sitemap which causes an index
file to be created to reference multiple sitemaps

2. Added back the conditional to store sitemaps only if env==production.
Not sure why we had commented that out.

3. Added a second generator to create another sitemap with the filename
'agencies' to house urls for the agency show pages. I added the filename
so not to overwrite our current urls

In the end, though I think we want to include this new content in this
way, I decided not to push because of the small number of agencies we
have, the fact that search engines will eventually find them organically
since they are all the subject of internal links and because I don't
want to negatively affect our growing presence in search results.

### DIFF
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -5,27 +5,38 @@ SitemapGenerator::Sitemap.default_host = "http://ebwiki.org"
 SitemapGenerator::Sitemap.public_path = 'tmp/'
 
 SitemapGenerator::Sitemap.sitemaps_path = 'sitemaps/'
+SitemapGenerator::Sitemap.create_index = true
 
-# if Rails.env.production?
+if Rails.env.production?
 # Instance of `SitemapGenerator::s3Adapter -- could have used wave adapter instead
-SitemapGenerator::Sitemap.adapter = SitemapGenerator::S3Adapter.new(
-  fog_provider: 'AWS', 
-  aws_access_key_id: ENV['AWS_ACCESS_KEY_ID'],
-  aws_secret_access_key: ENV['AWS_SECRET_KEY_ID'],
-  fog_directory: ENV['FOG_DIRECTORY'],
-  fog_region: ENV['S3_REGION'])
+  SitemapGenerator::Sitemap.adapter = SitemapGenerator::S3Adapter.new(
+    fog_provider: 'AWS',
+    aws_access_key_id: ENV['AWS_ACCESS_KEY_ID'],
+    aws_secret_access_key: ENV['AWS_SECRET_KEY_ID'],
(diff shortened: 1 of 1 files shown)


campgurus:
Right now, none of these agency pages seem to be indexed tho I have only checked a few.  For now, I can submit the agencies/index page and try to get Google to crawl all of the linked pages.

Going forward, having the sitemap which gets submitted automatically will be helpful.

trystant:
There are a couple of things:

1. First & foremost, I'm not clear on why those are in two separate sitemap files. You can add both of those to the same index; just put both article & agency loops in the same sitemap file. The sitemap should be able to handle both articles and agency pages.

2. I can see the production conditional being used in order to prevent staging.ebwiki.org sitemaps from being indexed, since they're on the same domain. I wonder if there's a way to set up a different domain for staging objects.

campgurus:
Ultimately sitemaps are limited by Google to < 50,000 urls. We currently know of 12,000 cases. for each case, we submit 2 pages currently for the show page and case followers page.  Both of those are currently within the articles sitemap. Later we might decide to submit other article-related pages such as articles/john-doe/history to the same sitemap. If we do, we will quickly approach the limit and need a new sitemap anyway. 

Mainly, I think it would be better practice to have a separate sitemap files for pages relating to agencies, events, organizations and potentially other categories, whether it is for better indexing by search engines or isolating potential sitemap issues.

trystant:
We currently have ~2300 cases.  We're not going to make it to 50,000 for awhile at this point.

The followers page isn't as important. I could see that being removed from the index.

I do see how using separate sitemaps can get over the 50,000 URL limit, but I think that it's a premature optimization, especially since it's a relatively simple adjustment to make.

campgurus:
I am all for not crossing bridges until we reach them. However, in this case, there are 18,000 law enforcement agencies in the US. Also, this article makes a case for doing multiple sitemaps sooner vs later: https://moz.com/blog/multiple-xml-sitemaps-increased-indexation-and-traffic

If you still don't agree, we can drop it.

trystant:
BTW, I've been convinced on this. Let's roll with the multiple sitemaps.

I  still don't think there's an answer about staging, however

campgurus:
I don't think we should have sitemaps for staging, since we don't want those pages to get indexed. Can we go ahead and merge this?

Issue #881, opened 2017-04-07
Geolocation is not working


campgurus:
I added several cases today and despite entering valid addresses, no latitude/longitude coordinates were generated. I also looked back at a few recent cases and it seems that this issue has existed for some time now, since they were also missing the geolocation.

trystant:
We need to add tests to confirm that geo works as well as add a rake task to autogenerate coordinates for all articles that don't have it (so that if this does happen again, the fix will take a few minutes).

campgurus:
We have a test that purports to test for geocoding of new articles within the model spec article_spec.rb

rlgreen91:
In looking at the test, I see that we check that the article has geocoding - but I can't find where we actually do that.  In `article.rb`, we include `after_validation :geocode` but I can't find any method definition associated with that.  I'll check again, but is there something I'm missing?

campgurus:
I think that method is in the geocoder gem.  https://github.com/alexreisner/geocoder

rlgreen91:
This looks like it is addressed in merged PRs.

PR #896, opened 2017-05-17
Address Geolocation issues

IsaacBell:
This was written to resolve issue #881  . Please check it out and let me know what you think.

All worked as intended from what I can tell  on my machines; let me know if that's not the case for others. I added a slight modification of the rake task from the geocoder docs for the back-generation of article coordinates. The command is to run that is:
`rake geocode_data:all CLASS=Article SLEEP=0.25 BATCH=100`

### DIFF
--- a/app/models/agency.rb
+++ b/app/models/agency.rb
@@ -9,6 +9,22 @@ class Agency < ActiveRecord::Base
 
   extend FriendlyId
   friendly_id :slug_candidates, use: :slugged
+  
+  scope :by_state, -> (state_id) {where(state_id: state_id)}
+  
+  # Geocoding
+  geocoded_by :full_address
+  before_save :geocode, if: Proc.new {|agcy| 
+    agcy.street_address_changed? || agcy.city_changed? || agcy.state_id_changed? || agcy.zipcode_changed?
+  } # auto-fetch coordinates
+  
+  def full_address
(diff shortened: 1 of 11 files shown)


trystant:
Hey @IsaacBell , it looks like the [Travis](https://travis-ci.org/BOWiki/BOW/builds/233431544) build failed for the PR. Considering that `master` is still green, would you mind looking into why the build is failing on your PR?  Thanks.

IsaacBell:
Whoops, thought I'd commented already. Same issue with AWS environment
variables (aws_access_key_id, etc.) as pops up in the Issues for
development, because I don't have production key env variables set.

trystant:
I think that you may need to be a member of the organization, @IsaacBell. I've just sent an invite. Once you accept, Travis should run the build *with* the environmental variables that you need.

trystant:
@IsaacBell BTW, due to the way Github does permissions, you might need to reconnect your Github account to Travis in order for it to pick up the organization permissions.

trystant:
So here's some additional information. Due to the way that [Travis](https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions) handles pull requests from forks, the environmental variables will not get copied over to your fork, @IsaacBell. 

If I were you, I'd just work off of the BOWiki repository directly, and file your pull request from that repository. Does that make sense?

IsaacBell:
Yes, makes sense, I'll do that.

Issue #907, opened 2017-05-20
Alphabetize agencies in new case form


rlgreen91:
In the new case form, the list of agencies in the agencies field should be alphabetized to make it easier to select the relevant agencies.

rlgreen91:
This looks like it's sort of done?  For some reason, it's as if multiple sets of data are being pulled, and each one is alphabetized, rather than one set being pulled and that data being alphabetized.  Will need to investigate more.

rlgreen91:
One particular issue - some of the agencies are being saved with a tab character or space at the beginning, causing them to be listed first.  Now going to investigate why that's happening.

rlgreen91:
I've gone through and edited the agencies to remove the tabs and spaces at the beginning of the names.  This lead to a couple of duplicates being combined.  Now to see how that could happen and at the least strip any beginning spaces.

rlgreen91:
Last part of this - making sure numbers are handled properly. For example, 1st should come before 3rd and 18th.

rlgreen91:
It could be.  I have a game plan though.

Issue #903, opened 2017-05-20
ActiveRecord::SubclassNotFound: The single-table inheritance mechanism failed to locate the subclass: 'local'. This error is raised because the column 'type' is reserved for storing the class in case of inheritance. Please rename this column if you didn't


trystant:
View details in Rollbar: [https://rollbar.com/trystant/BOW/items/691/](https://rollbar.com/trystant/BOW/items/691/)

```

NameError: wrong constant name local
  File "/usr/local/rvm/gems/ruby-2.3.0@ebwiki/gems/activesupport-4.2.7.1/lib/active_support/inflector/methods.rb", line 261, in const_get
  File "/usr/local/rvm/gems/ruby-2.3.0@ebwiki/gems/activesupport-4.2.7.1/lib/active_support/inflector/methods.rb", line 261, in block in constantize
  File "/usr/local/rvm/gems/ruby-2.3.0@ebwiki/gems/activesupport-4.2.7.1/lib/active_support/inflector/methods.rb", line 259, in each
  File "/usr/local/rvm/gems/ruby-2.3.0@ebwiki/gems/activesupport-4.2.7.1/lib/active_support/inflector/methods.rb", line 259, in inject
  File "/usr/local/rvm/gems/ruby-2.3.0@ebwiki/gems/activesupport-4.2.7.1/lib/active_support/inflector/methods.rb", line 259, in constantize
  File "/usr/local/rvm/gems/ruby-2.3.0@ebwiki/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb", line 566, in get
  File "/usr/local/rvm/gems/ruby-2.3.0@ebwiki/gems/activesupport-4.2.7.1/lib/active_support/dependencies.rb", line 597, in constantize
  File "/usr/local/rvm/gems/ruby-2.3.0@ebwiki/gems/activerecord-4.2.7.1/lib/active_record/inheritance.rb", line 181, in find_sti_class
  File "/usr/local/rvm/gems/ruby-2.3.0@ebwiki/gems/activerecord-4.2.7.1/lib/active_record/inheritance.rb", line 169, in discriminate_class_for_record
  File "/usr/local/rvm/gems/ruby-2.3.0@ebwiki/gems/activerecord-4.2.7.1/lib/active_record/persistence.rb", line 67, in instantiate
[...]

Issue #853, opened 2016-12-27
Add clearer error message to create agency form


campgurus:
Error message says to fix errors below, but it is not clear what field it refers to.

trystant:
So here is the screenshot that I got for the new agencies page:
![screenshot 2017-04-02 at 4 53 08 am](https://cloud.githubusercontent.com/assets/13509/24585776/54abfb80-1760-11e7-9ce3-6b24c228edcf.png)

Since the only validation on agencies is the following:
```
class Agency < ActiveRecord::Base
...

  validates :name, presence: true
  validates :name, uniqueness: true
  validates :state_id, presence: true
```

It makes sense that the name would be highlighted. At the same time, there doesn't seem to be an error message for the state id, which could cause some confusion.

rlgreen91:
@trystant Please put an asterisk next to the state dropdown as well to indicate it's required.

campgurus:
i am surprised rails did not add the asterisk automatically based on the presence validation...

IsaacBell:
We can add a notifier at the top as well to increase clarity. Screenshot for example: 

![screenshot 2017-05-19 at 1 15 53 pm](https://cloud.githubusercontent.com/assets/2613157/26261086/66ada77a-3c95-11e7-939b-52354a30e4f9.png)

trystant:
@IsaacBell updating the 'message' part of the validation, you're saying?

IsaacBell:
@trystant that too, but in this screenshot the errors are listed at the top ("2 errors prohibited this article from being saved") in addition to being highlighted red in their respective fields.

Could use some style cleanup on the error box as well, but I wanted to throw the idea out first.

trystant:
I like where this is going. I say full speed ahead

rlgreen91:
In the error box, please put a colon after the field name.  So it looks like "Name: Agency names must be unique.  This agency already exists." for example.

Also, for the name message, it might be better to say something like this:
"Name: An agency with this name already exists and can be found <u>here</u>.  If you want to create a new agency, it must have a unique name."
The underline is meant to indicate that it's a link - it would be nice to link to the existing agency's page, so that if the user is looking to edit the agency, they can just click that and go straight to the page.  Of course, I'm assuming that someone gets to the page with the sole intention of adding a new agency - if they're doing this in the process of some larger goal, like adding a new case, then we might change the behavior.

Also, in the state message, it might be better to say something like this:
"State: You must specify the state in which the incident occurred."

[...]

Issue #738, opened 2016-07-26
Add type to Agency model


campgurus:
need to distinguish between, state, local, federal, university and private law enforcement agencies.

rlgreen91:
Addressed in PR #908.

PR #908, opened 2017-05-20
Agency Workflow

IsaacBell:
Issue resolution and refactoring.

Added a 'jurisdiction' field to the Agency model in order to fill the function of #738 . It shows as a dropdown in the form. (Calling it 'Type' gave me errors, see #903 ) 

Should this field be labelled "type", "jurisdiction", or something else in the view?

Other issues addressed: #907 ,  #853 ,  #901 .

### DIFF
--- a/app/views/agencies/show.html.erb
+++ b/app/views/agencies/show.html.erb
@@ -5,9 +5,12 @@
 
 <span  class="page-header articleTitle">
   <h1 class="page-header articleTitle"><%= @agency.name %></h1>
-  <h4><%= @agency.street_address %></h4>
-  <h4><%= @agency.city %>, <%= State.find(@agency.state_id).name %> <%= @agency.zipcode %></h4>
+  <h4> <%= @agency.full_address %> </h4>
 </span>
+
+<p>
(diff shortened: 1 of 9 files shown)


app/views/agencies/show.html.erb
```diff
@@ -5,11 +5,14 @@
 
 <span  class="page-header articleTitle">
   <h1 class="page-header articleTitle"><%= @agency.name %></h1>
-  <h4><%= @agency.street_address %></h4>
-  <h4><%= @agency.city %>, <%= State.find(@agency.state_id).name %> <%= @agency.zipcode %></h4>
```
trystant: Hey, I know this has already been submitted, but can we turn this into a scope on `State` and use that in the view instead? Thanks.

IsaacBell: Sounds simple, but I need more clarification. Would you like for me to have two h4's as in the old code, and replace the State name evaluation with a scope?

trystant: Yeah, it's pretty simple, just replace
` <%= State.find(@agency.state_id).name %> `
with a scope perhaps named
` <%= State.from_agency(@agency).name %>

Issue #909, opened 2017-05-21
Escape the html in the Maps flyover descriptions


IsaacBell:
Example: https://ebwiki.org/articles/emanuel-jean-baptiste

![screenshot 2017-05-20 at 10 50 46 pm](https://cloud.githubusercontent.com/assets/2613157/26281035/f5b59ed6-3dae-11e7-8d98-441409e4420d.png)

rlgreen91:
This has been addressed in #1593, #1623, #1627 and #1629

PR #912, opened 2017-05-21
#909 Escape the html in the Maps flyover descriptions

IsaacBell:


IsaacBell:
screenshot 2017-05-21 at 12 44 07 am

Issue #927, opened 2017-05-28
Refactor if-else statement in welcome email view


rlgreen91:
In views/user_notifier/welcome_email.html.erb, refactor the following code such that only the text about being a newsletter subscriber is within an if-else statement.

```
<% if @user.subscribed %>
    As a newsletter subscriber, you'll receive our general updates periodically.
    <% if @user.all_following.count <= 0 %>
      Likewise, it is very important that you click to follow specific cases and allow us to keep you up to date. The more people paying attention, the easier it will be effect change.</p>
    <% else %>
      Likewise, you have already taken the first step by following <%= pluralize @user.all_following.count, "case" %> on EBWiki.  Thanks for allowing us to keep you up to date!
    <% end %>
  <% else %>
    <% if @user.all_following.count <= 0 %>
		  It is very important that you click to follow one or more cases and allow us to keep you up to date. The more people paying attention, the easier it will be effect change.
		<% else %>
		  You have already taken the first step by following <%= pluralize @user.all_following.count, "case" %> on EBWiki and allowing us to keep you up to date. <%= link_to 'Subscribe to our newsletter as well', ENV['MAILCHIMP_LINK'] %> for periodic general updates and commentaries on this issue.
		<% end %>
  <% end %>
```

rlgreen91:
Addressed in #1470

Issue #926, opened 2017-05-28
Remove logic from new notification email views


rlgreen91:
Remove logic from views/mailboxer/notification_mailer/new_notification_email.html.erb:

`<h1>You have a new notification: <%= @notification.subject.html_safe? ? @notification.subject : strip_tags(@notification.subject) %></h1>`

Remove this logic from views/mailboxer/notification_mailer/new_notification_email.text.erb:

`<%= @notification.subject.html_safe? ? @notification.subject : strip_tags(@notification.subject) %>`

`<%= @notification.body.html_safe? ? @notification.body : strip_tags(@notification.body) %>`

These should be refactored into a helper.

Issue #925, opened 2017-05-28
Remove logic from layouts header partial


rlgreen91:
Remove the following logic from views/layouts/_header.html.erb:

```
<div class="navbar-form navbar-left">
          <%= select_tag(:state_id, 
            options_from_collection_for_select(State.all, :id, :name), 
            include_blank: "Filter By State",
            class: 'form-control select2') %>
        </div>
```

Honestly, we should have a set of helpers just for generating a list of states, a list of state ids, etc.

Issue #924, opened 2017-05-28
Remove logic from kaminari page partial


rlgreen91:
Remove logic from views/kaminari/_page.html.erb:

```
<%= content_tag :a, page, remote: remote, rel: (page.next? ? 'next' : (page.prev? ? 'prev' : nil)) %>

<%= link_to page, url, remote: remote, rel: (page.next? ? 'next' : (page.prev? ? 'prev' : nil)) %>
```

These lines should be refactored using a helper.

PR #930, opened 2017-06-02
Refactoring

IsaacBell:
This batch of fixes is relevant to the following Issues:

#927 
#919 
#925 
#926 
#920 
#922 
#907 
#912 
#918 
#930 
#923 
#915 
#924 

Please let me know if you see any issues here, would like alterations, or want me to generally alter syntax or coding style anywere. Otherwise I do not plan on merging this PR myself manually without feedback, because I want to ensure my code is lining up to expectations and style first. Feel free to reach out directly or post up  questions/concerns/feedback.

### DIFF
--- a/app/views/articles/_form.html.erb
+++ b/app/views/articles/_form.html.erb
@@ -81,7 +81,7 @@
   <fieldset class="fsStyle">
     <legend class="legendStyle">Enforcement Agencies Involved <%= image_tag('help_icon.png', data: {:toggle => 'popover', :container => 'body', :trigger => 'hover',:placement => "bottom", :title => "Agencies", :content => "Add law enforcement agencies involved here so we can grade their performance."})%></legend>
     <div id="links">
-      <%= agencies_list %>
+      <%= f.association :agencies, collection: agency_dropdown_collection, input_html: { multiple: true, class: 'select2'} %>
     </div>
 
     <% if params[:action] == "edit" %>
--- a/app/views/articles/_form.html.erb
+++ b/app/views/articles/_form.html.erb
@@ -81,10 +81,7 @@
   <fieldset class="fsStyle">
     <legend class="legendStyle">Enforcement Agencies Involved <%= image_tag('help_icon.png', data: {:toggle => 'popover', :container => 'body', :trigger => 'hover',:placement => "bottom", :title => "Agencies", :content => "Add law enforcement agencies involved here so we can grade their performance."})%></legend>
     <div id="links">
-      <%= f.association :agencies, 
-      		collection: Agency.all.sort_by{ |e| ActiveSupport::Inflector.transliterate(e.name.downcase) }, 
-      		input_html: { multiple: true, class: 'select2'}
-      %>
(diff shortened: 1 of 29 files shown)


app/views/articles/_form.html.erb
```diff
@@ -81,10 +81,7 @@
   <fieldset class="fsStyle">
     <legend class="legendStyle">Enforcement Agencies Involved <%= image_tag('help_icon.png', data: {:toggle => 'popover', :container => 'body', :trigger => 'hover',:placement => "bottom", :title => "Agencies", :content => "Add law enforcement agencies involved here so we can grade their performance."})%></legend>
     <div id="links">
-      <%= f.association :agencies, 
-      		collection: Agency.all.sort_by{ |e| ActiveSupport::Inflector.transliterate(e.name.downcase) }, 
-      		input_html: { multiple: true, class: 'select2'}
-      %>
+      <%= f.association :agencies, collection: agency_dropdown_collection, input_html: { multiple: true, class: 'select2'} %>
```
rlgreen91: LGTM

IsaacBell:
Further changes made based on feedback and other issues addressed

spec/helpers/articles_helper_spec.rb
```diff
@@ -42,5 +42,25 @@
       expect(helper.article_sanity_check).to be_falsey
     end
   end
+
```
trystant: These specs generally make sense, but what about failure cases? What happens if Article.recent(10) returns nil?

IsaacBell: I have to turn that question back around. What _should_ happen in that case? If there aren't ten recent articles to pull from, nothing should happen and no error should be thrown to my knowledge as the ActiveRecord query will return [] as opposed to nil. 

We could print a message if no articles are found ("No cases added yet"). 

Am I on track?

trystant: the behavior is correct and the helper should probably return an empty array. Returning a string in case nothing is found can work; the specs would need to capture that expectation.

IsaacBell: Added specs to address this

campgurus:
just curious as to the status of this PR.

trystant:
@IsaacBell and i were going through rounds of edits on this, and then his focus changed due to a new job. It would probably be good for us to check in with each other on this again.

rlgreen91:
I had comments on this - new comments that needed to be addressed.  I know it might be too late at this point @trystant, but I'd really prefer for this to be broken up into smaller PRs.  Like, one per issue if possible.

rlgreen91:
Anyways, I'll go back and add my comments again apparently.

trystant:
@rlgreen91 you're right. The smaller PRs would make more sense. Easier to be certain about scope & items delivered.

rlgreen91:
All of the changes here have been replicated in other PRs that have been merged, and all of those corresponding issues have been closed.  With that, we can close this PR.

PR #985, opened 2017-08-23
Adds automatic check via bullet to test environment.

rlgreen91:
Configures Bullet to run for each test, both in a local testing environment as well as in Travis when pull requests are opened.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,12 @@ After the Walter Scott video was released showing the vast discrepancy between t
 
 ## Testing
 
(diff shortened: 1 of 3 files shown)


rlgreen91:
@trystant Merge please?

trystant:
Hey, just to be clear, how is the Travis build affected if there is an N+1 query?

rlgreen91:
When Bullet detects an N+1 query, we've set it to raise an exception, which means that the test will fail and the Travis build will fail.   Additionally, Bullet is set to raise that exception for tests period, so if someone is testing locally and there is an N+1 query, it will fail locally as well.

PR #1000, opened 2017-08-26
870 rename notify of removal

rlgreen91:


trystant:
Hey @rlgreen91 , have you used your credientials recently and confirmed that they worked independently? I recall you mentioning that you were using different IAM credientials for this S3 access. Could there be something about that change that is affecting the `fog` configuration?

rlgreen91:
Ok, so this branch had issues, mostly due to configuration problems on my end.  I have set up a new dev environment, made these changes there, and will create a PR from that.  So now, this one can die.

Issue #959, opened 2017-07-27
Add spacing on individual case pages


rlgreen91:
On the show view for an article, the related cases section touches the map.  Add some space between the two.

PR #1068, opened 2017-10-14
issue #959: added space between related cases and map

dpborland:
Something like this?

screen shot 2017-10-14 at 2 29 01 pm


rlgreen91:
@dpborland - yes, just like that. Once CI finishes I'll merge it.

Issue #1110, opened 2017-11-18
Remove numbers in table on agencies index page


rlgreen91:
On the agencies index page, remove the first column of the table.  This should also allow you to use `each` rather than `each_with_index` and to truncate after 50 characters, rather than 30.

Issue #915, opened 2017-05-27
Remove logic from agencies index page


rlgreen91:
Remove the following logic from `view/agencies/index.html.erb:

`<% Agency.all.order(:name).each_with_index do |agency, i| %>
    <tr>
      <td><%= i + 1 %></td>
      <td><%= link_to truncate(agency.name, length: 30), agency %></td>
      <td><small><%= agency.updated_at.strftime("%m.%e, %l:%M %p") %></small></td>
      <td><%= link_to 'Edit', edit_agency_path(agency) %></td>
    </tr>
  <% end %>`

It's probably easiest to just stick this in a helper rather than do something controller side.

PR #1124, opened 2017-11-29
Updates seeds file and removes logic from agencies index view

rlgreen91:
Updates the seed file to create agencies.  Also removes logic from the agencies index view and instead creates functions within `app/helpers/agencies_helper.rb`.

This PR directly addresses issues #1110 and #915.


rlgreen91:
The code climate issues are non-issues, so that's good to go as far as this is concerned.

Issue #918, opened 2017-05-27
Remove logic from analytics index page


rlgreen91:
Remove the following logic from views/analytics/index.html.erb:

```
<div class="chart-stage">
          <%= column_chart @visits.group_by_day(:started_at).count %>
</div>
```

```
<div class="chart-stage">
          <div id=""><%= pie_chart @visits.where(started_at: 1.days.ago..Time.now).group(:browser).count %></div>
</div>
```

```
<div class="chart-stage">
          <%= column_chart @users.group_by_day(:created_at).count %>
</div>
```

```
<div class="chart-stage">
          <div id=""><%= pie_chart @visits.where(started_at: 1.day.ago..Time.now).group(:referring_domain).count %></div>
        </div>
```

```
<% Article.order(updated_at: :desc).first(10).each_with_index do |article, i| %>
             <tr>
               <td><%= i + 1 %></td>
               <td><%= link_to truncate(article.title, length: 14), article %></td>
               <td><small><%= article.updated_at.strftime("%m.%e, %l:%M %p") %></small></td>
             </tr>
           <% end %>
```

```
<% Article.order(follows_count: :desc).first(10).each_with_index do |article, i| %>
             <tr>
               <td><%= i + 1 %></td>
               <td><%= link_to truncate(article.title, length: 25), article %></td>
               <td><%= link_to article.follows_count, articles_followers_path(article) %></td>
             </tr>
           <% end %>
```

```
 <% Visit.where(started_at: 7.days.ago..Time.now).group(:landing_page).order('count_id DESC').limit(13).count(:id).each_with_index do |visit, i| %>
            <tr>
              <% if Article.find_by_slug(visit[0].split('/').last).present? %>
                <td><%= i %></td>
                <td><%= link_to truncate(Article.find_by_slug(visit[0].split('/').last).title, length: 25), article_path(Article.find_by_slug(visit[0].split('/').last)) %></td>
                <td><%= visit[1] %></td>
              <% end %>
            </tr>
           <% end %>
```

```
<% Comment.order(created_at: :desc).each do |comment| %>
            <li>
              <blockquote class="blockquote">
                <p><%= link_to truncate(Article.find(comment.commentable_id).title, length: 30), article_path(Article.find(comment.commentable_id)) %> -- <%= comment.created_at.strftime("%B %d, %Y") %></p>
                <p><%= link_to truncate(comment.content, length: 100), article_path(Article.find(comment.commentable_id)) %></p>
[...]

PR #1138, opened 2017-12-06
Refactors analytics index page

rlgreen91:
This PR refactors the analytics index page by moving logic from the view to a helper, and adding scopes and service objects to reduce logic and queries.  Noe that there are no tests for the helpers due to the fact that all of the added methods are merely wrappers for Rails and Ruby methods.

Note that, even using a database dump, it is not possible to view the page in question locally due to the lack of data.  We'll have to proof this in staging.

This PR addresses Issue #918


rlgreen91:
This PR is updated and replaced by PR #1260.

PR #1201, opened 2017-12-26
1099 redo editors list

campgurus:
this is a redo of the previous 1099_editors_list PR which I will close.


app/views/analytics/index.html.erb
```diff
@@ -153,19 +147,50 @@
...
+        </div>
+        <div class="chart-stage">
+         <table class='table leader-list'>
+           <tr>
+            <th>Editor</th>
+            <th>Updates</th>
+           </tr>
+
+           <% case_updaters_count(10).each do |who| %>
+            <tr>
```
rlgreen91: It might be useful to add a couple of more lines before `<tr>` that assign the different elements of `who` to more distinct variables.  So something like:

`
<% case_updaters_count(10).each do |who| %>
<% user_id = who[0] %>
<% user_object = who[1] %>
<tr>
`

rlgreen91:
Also, I'm not sure why production dumps and backups are showing up in your PR.  If there were no changes, remove them from the PR.

campgurus:
Can we merge this?  more changes are being added to master s.t. I am concerned that we will have unmergeable conflicts and I will need to manually rebase each separate commit again.

rlgreen91:
No, we shouldn't merge this - we shouldn't be querying the database in the view.  When you do `Article.order.` and stuff like that, that needs to be called at the controller level and passed as an instance variable.  All that should be happening in the view is something like `@articles.each do`.

Also, you don't need to rebase after each commit. When you get to a stopping point and realize you need to merge in changes from master, you only need to rebase once, and git will add your changes on top, regardless of whether you have one commit or 100.

campgurus:
Ok so I will go back in and clean these db calls.  Just to clarify, in the top_metrics partial, we call eg Follow.count.  Should that be abstracted with an instance variable in the controller, like @follows=Follow.all and then in the view, use @follows.count?

Should I create separate helpers for every iteration, like def total_follows_count, def total_distinct_followers, def recently_updated_cases, etc?

How do you guys want me to do this?

campgurus:
What is the status of this PR?

rlgreen91:
Hey, so I think this looks good for right now.  Some of these things might change as we continue to refactor that page in general.  Feel free to merge whenever, unless @msnyon has any objections.

I missed your earlier question, but here's my thought - I personally think that DB queries shouldn't be in the view layer at all.  Like, it should be a scope or method called by the controller and passed as an instance variable.  The reasons why I think we can just give this one a pass is because I know I'm going to refactor it - I'm still going through the other views and removing logic, and since I'm doing most of them, I can see a few common methods we can generalize that should simplify this and some other files.  Another reason I think we're good is that I have an idea of what we want to do next in terms of the analytics page, so some of this stuff will probably move.  Even just having this in the view helper as opposed to the view will make it easier to do that.

rlgreen91:
@campgurus What's the blocker on this?

rlgreen91:
@campgurus What's the blocker on this?

campgurus:
Not sure. We were good and ready to merge a while back, but now there are conflicts based on later changes to analytics views. Aside from manually redoing all of the changes in this PR, what needs to happen?  Does this need to get rebased on top of current master?

rlgreen91:
@campgurus Yes, you do need to pull the latest version of master and rebase on top of that.  Additionally, you need to remove the two binary files that got added to this PR.

You'll also want to check the methods you added in `analytics_helper` - I tried my best to not have conflicts, but I think a lot of the queries you're making are part of scopes now, or otherwise at the model level.

If this is a ton to do for this PR, then just let me know and I'll open another issue to reproduce the changes in this PR.

campgurus:
Hey @rlgreen91 do you mind making a new issue and assigning it to me? probably easier to just start on top of the up to date master with this minor addition of the editors list.

rlgreen91:
@campgurus I can, but are you sure you don't want one of the interns to do it?

campgurus:
That is fine. the coding for the Editors list is done if the look at the first commit in this PR.

Issue #1096, opened 2017-11-07
Highlight content on the landing page.


rlgreen91:
On the landing page, bold the number in the headline, i.e. "These XXX people..."

Related to Issue #845

campgurus:
here is a screenshot of the current styling of the metric on the home page:

<img width="781" alt="screen shot 2018-01-07 at 7 13 45 am" src="https://user-images.githubusercontent.com/1260659/34649295-66658ebc-f37a-11e7-8358-3ea90c5089ce.png">

rlgreen91:
I think this image is supposed to be in the PR.  I'll give my thoughts there.

rlgreen91:
Now, to complete this issue, you need to change the number to be the orange used throughout the site.  You don't have to make it bold, just change the color.

rlgreen91:
Addressed in #3255

PR #1242, opened 2018-01-07
for #1096 highlighting landing page metric

campgurus:
i took a shot at highlighting the text by increasing the size slightly, changing the background color and adding a thin outline to the text area.

I branched from the current master, so it still contains the failing test relating to the analytics page metric we have been discussing recently.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -126,3 +126,12 @@ ul {
 .select2-container--classic .select2-selection--single {
   filter: unset;
 }
+
+/* highlight certain text such as number of cases on home page */
+
+.highlight-metric {
+  background-color:#E0661D;
+  font-size: 1.25em;
+  color: #555;
+  outline: #747474 solid 2px;
+}
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -130,8 +130,6 @@ ul {
 /* highlight certain text such as number of cases on home page */
 
 .highlight-metric {
-  background-color:#E0661D;
-  font-size: 1.25em;
-  color: #555;
-  outline: #747474 solid 2px;
+  font-size: 1.5em;
+  color: #E0661D;
 }
--- a/app/views/articles/index.html.erb
+++ b/app/views/articles/index.html.erb
@@ -7,7 +7,7 @@ meta title: "EBWiki - Ending Violence Against People of Color",
 <div class="container">
     <div class="row">
         <div class="col-lg-9">
-            <h2 class="page-header">These <%= Article.count %> people were killed by police.<br>We track their cases here.</h2>
+            <h2 class="page-header">These <span class="highlight-metric"><%= Article.count %></span> people were killed by police.<br>We track their cases here.</h2>
 
             <div class="text-center">
               <h3>


rlgreen91:
I'm assuming that the screenshot added to Issue #1096 was meant to go here, so I'll add my thoughts on it here.

I don't like that - having a box seems odd, and that particular color combination makes it seems muddy - it's not a clean look.  I think it would be best to do two things: first, just make that number orange, and keep the rest of the text black.  Second, I would increase the font size of the number. I don't know what the font size of the line is, but if it's, for example, 12pt, then try seeing what it looks like with the number at 16pt or 18pt.  I think that'll do more to provide the contrast users wanted to see.

rlgreen91:
Ah, also - you may need to rebase since I merged your other PR.

campgurus:
Screenshot:

screen shot 2018-01-14 at 8 02 38 am

Note, the larger metric pushes the thin line separating the header and the photo gallery down a bit so it no longer lines up with the line beneath "Recently Updated."

Incidentally, I am thinking we may want to move the recently updated list to the bottom of the page. Although we need to add a class to the links in recently updated list so when we review ahoy data, we can see if users are clicking on the links in that list.

rlgreen91:
Now that you mention that line, I'm kind of against it.  I don't really see a need for having a line there - it's not necessary to have separation between the headline and the buttons, just white space will do that fine.

I know in the file that you increased font-size to 1.5em, but I'd be interested to hear your opinion for larger sizes, like 2 or 2.5em.  Did it look odd?

I wonder how this would look if, rather than aligning the number along the bottom of the text, we instead centered it.  What I mean is that right now, the bottom of each character in the headline is aligned vertically.  However, the number sticks out in a sense because it's bigger - it looks unbalanced.  I'm wondering how it would look if the number was aligned with the text such that it was centered vertically based on the bottom of the characters in the text.  That might help it look more balanced while still drawing attention.

I personally think the Recently Updated section should stay on the right since it draws people to what are probably the most popular cases as the time.  Don't want to hide it or annoy visitors by making them scroll.

campgurus:
This is 2.5 em:

screen shot 2018-01-15 at 6 06 57 pm

this is 2.5 em centered:

screen shot 2018-01-15 at 6 18 58 pm

I think the first one looks better.  And if we reword to make get rid of "These" and make the number first it might also help.

screen shot 2018-01-15 at 6 31 07 pm

rlgreen91:
I agree with getting rid of "these".  How about this - we put the number to the left of all of the text, and then put the text on the two lines in one block?  Basically just aligning the bottom line of text with the text on the top, and enlarging the number to span the height of both lines.

rlgreen91:
@campgurus What's the blocker on this?

rlgreen91:
@campgurus What's up with this issue?  It looks like we were still waiting to see what a couple of ideas would look like.

rlgreen91:
@campgurus have you had a chance to look at this?

rlgreen91:
@campgurus what's going on with this?

campgurus:
just looking back on this issue. I completely spaced on this.  I think it is probably better for someone who has a better facility with css and design esp if we are redesigning the entire page.

trystant:
Based on the previous feedback, closing this.

PR #1260, opened 2018-01-22
Refactors analytics index page

rlgreen91:
There is a lot going on in this PR: as a brief overview, it refactors the analytics index page, adds a service object to determines visits to articles, and adds the newest production database dump.

In more detail: this PR contains heavy refactoring of the analytics index page, basically moving all logic out of the view - now, at best, the view loops through a collection of articles, visits, etc.  There were quite a few scopes added to the various models, and those are called within the controller and passed as instance variables.  There are also helper methods used to clean up some of the code related to presentation - note that these methods are untested, as they are mostly cleaner interfaces for one-liner Rails methods.  Additionally, the column numbers in several of the tables of the analytics page were removed - a screenshot is included in a comment.

This PR also contains a new service object that, when given the information about views to URLs, will return a list of articles those URLs lead to and the number of views.

Finally, the PR also includes the latest production database dump and removes the two previous ones.


rlgreen91:
analytics_screenshot_1_22_2018

rlgreen91:
Two more things: one, this addresses Issue #918.  Second, this is meant to be a replacement for a previous PR which was seriously outdated and originated on a laptop that can no longer connect to the internet.

Issue #922, opened 2017-05-28
Remove logic from articles index page


rlgreen91:
Remove the following logic from views/articles/index.html.erb:

```
<% Article.order(updated_at: :desc).first(10).each do |article| %>
               <tr>
                 <td><%= Article.order(updated_at: :desc).first(10).index(article) + 1 %></td>
                 <td><%= link_to truncate(article.title, length: 14), article %></td>
                 <td><small><%= article.updated_at.strftime("%m.%e, %l:%M %p") %></small></td>
               </tr>
             <% end %>
```

PR #1263, opened 2018-01-27
Refactors recently updated cases section in cases view.

rlgreen91:
This PR refactors the recently updated cases section in `app/views/articles/index.html.erb`.  Specifically, the database query is performed in the corresponding controller action, and the number column is removed.

This PR addresses Issue #922.

### DIFF
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -11,6 +11,7 @@ class ArticlesController < ApplicationController
 
   def index
     page_size = 12
+    @recently_updated_articles = Article.sorted_by_update 10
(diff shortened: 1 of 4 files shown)


rlgreen91:
recently_updated_cases_screenshot_1_26_2018

rlgreen91:
I'll be honest, I'm not understanding how that spec works.  And maybe this is due to a lack of understanding kaminari, but I could not get it to work if I used `FactoryBot.create`, just `FactoryBot.build`.  Which is fine, except my understanding is that `build` only returns an unsaved instance of the record.  How then does calling a scope on that same model return actual data, instead of nil? Or am I misunderstanding how scopes work in tests?  Would appreciate it if you could slide through @trystant and lend your thoughts.

PR #1295, opened 2018-02-24
refactor new method

StlMaris123:


StlMaris123:
I had open this pr to check whether Travis will work

Issue #1120, opened 2017-11-28
Show page buttons misaligned


campgurus:
beneath the photo, there are social share icons and three buttons. the icons look like they have spread out somehow.  likewise, the buttons are now also taking up the full width of the div and pressing against the content sections awkwardly. Screesnshot below:

<img width="375" alt="screen shot 2017-11-28 at 6 08 04 pm" src="https://user-images.githubusercontent.com/1260659/33349347-a7988826-d467-11e7-8acf-4da3d015538b.png">

rlgreen91:
Hmm, this is probably better addressed as an extension of issue #962 - this should all be in an aside element, which should fix the styling (or make it easier for you to do so).

ArushiSinghal:
Hello @rlgreen91 
https://github.com/EBWiki/EBWiki/blob/bb9f489e491f2024d2685a82d29f3486ec83ab05/app/assets/stylesheets/articles.scss#L78-L80
@simransinghal and I tried to fix this issue and found that width defines in this class is overwritten by the width:100%.
The reason is that it is overwritten by the bootstrap class.
So we are thinking of changing the line
https://github.com/EBWiki/EBWiki/blob/bb9f489e491f2024d2685a82d29f3486ec83ab05/app/views/articles/_button_group.html.erb#L1
to 
```<div class = "btn-group btn-group-justified" style="width:calc(100% - 30px)">``` .
Any suggestion from your side is very helpful or should we go ahead with this change.

**Comment by simransinghal** [comment]
Me and @ArushiSinghal will like to propose a solution for this. This issue can be resolved by using "!important".
That is by changing the code to:-
```
.btn-group-justified { 
     width: calc(100% - 30px) !important; 
     margin: 15px 0 15px 15px; 
} 
```
But, as using !important is not a good style. So, we will like to take others opinion before submitting PR.
cc @campgurus @rlgreen91

rlgreen91:
Hmm, I'd rather you not use `!important`.  I mentioned above that I would like for you to update the HTML to contain all of this in an `aside` element.  That might make it easier to style some of this.  

[...]

PR #1300, opened 2018-02-27
Fixed issue "page buttons are misaligned"

ArushiSinghal:
Issue #1120 
This pull request is to fix the page misalignment with the second suggestion as suggested by @simransinghal.

### DIFF
--- a/app/assets/stylesheets/articles.scss
+++ b/app/assets/stylesheets/articles.scss
@@ -75,7 +75,7 @@
 }
 
 .btn-group-justified {
-    width: calc(100% - 30px);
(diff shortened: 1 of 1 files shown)


rlgreen91:
Ok, so don't do this.  If you're unsure of what route to take on an issue, then sort it out on the issue.  Do not submit two PRs that amount to the same fix for one thing.  This can cause problems - a person might accept and merge the wrong PR, and close the one that should have been used.  Then, for projects like this, with multiple maintainers, you might have different people each merging one of the PRs, again causing confusion when the "wrong" code is merged.  In the future, just submit one PR.

Closing this PR.

PR #1331, opened 2018-03-03
Fixed control of access to users.

cam-ila:
This pull request fix control of access to users.
Any user could access by url and edit the information of another user.
Example: https://ebwiki.org/users/1/edit

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -162,4 +162,4 @@ gem 'rack-host-redirect'
 gem 'rubocop', require: false
 
 # for authorization system.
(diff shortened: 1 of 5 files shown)


rlgreen91:
@cam-ila Thanks for catching this!  While we are planning to add Pundit, we are going to add policies in a more systemic way.  I would advise closing this for now and working on one of the issues in the project post in Slack.  I'll add this back as an issue, I just need a bit of time to plan how we'll go about this.

PR #1453, opened 2018-04-10
create fixtures file for states fixes #1336

### LINKED ISSUE #1336: Create fixture file for states
Create a fixtures file `states.yml` containing each state.  Just copy, more or less, the info within `db/seeds.rb`.

rachaelkiteme:
In your PR did you:

### DIFF
--- a/spec/fixtures/states.yml
+++ b/spec/fixtures/states.yml
@@ -1,0 +1,191 @@
+one:
+  title: 'Alabama' 
+
+two:
+  title: 'laska'
+
+three:   
+  title: 'Arizona' 
+
+four:   
+  title: 'Arkansas' 
+
+five:  
+  title: 'California'
+
+six:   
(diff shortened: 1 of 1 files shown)


rlgreen91:
Hmm, it might be better to have the fixtures labelled as the actual state names, rather than numbers.  So `one` becomes `alabama`, `two` becomes `alaska`, and so on.  There you can keep them all lowercase and use underscores instead of spaces when needed.  That will provide more clarity later in the tests.

rachaelkiteme:
Okay thank you @rlgreen91 will work on that.


Regards
Rachael

rlgreen91:
@rachaelkiteme For the fixture labels, if there is a space between two words, replace them with an underscore.  For example, `new york` should become `new_york`.  Having a space won't work properly once we start using them in the tests.

PR #1558, opened 2018-05-20
Put in fix for case editing issue

trystant:
Overall, there was an error occurring rendering the 'edit' template after a failed update. These changes were placed to stop that error from occurring.

- app/controllers/cases_controller.rb - Added Category.all & State.all
  to the edit form so that the page renders successfully

- spec/controllers/cases_controller_spec.rb - Added specs to confirm that
  those required values are in the rendered template

Updates #1556, #1557

In your PR did you:

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -70,6 +70,8 @@ class CasesController < ApplicationController
       UserNotifier.send_followers_email(@this_case.followers, @this_case).deliver_now
       redirect_to @this_case
     else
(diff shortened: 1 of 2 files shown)


trystant:
There is one potential issue; the alls in the failure part of the edit method. I feel like there's a better way to bring those up, and if categories somehow blew up, those could be a problem. That's for another day, however.

rlgreen91:
Tbh, not only should we rename `categories` to be more specific, but those can also be replaced with locale files.

Issue #1443, opened 2018-04-05
Improve performance of query for case versions


rlgreen91:
Improve the performance of the query for versions of a case.  Currently, the cases controller has a method `history` that runs the following query to retrieve the different versions:

```
@case_history = @this_case.try(:versions).sort_by(&:created_at).reverse
```

Replace that with the following statement:

```
@case_history = @this_case.try(:versions).sort_by { |version| -version.created_at }
```

Also include a comment explaining that adding the `-` essentially performs the reverse sorting at the database level, rather than pulling all of the data back within the Rails application and performing a reverse there.

StlMaris123:
@rlgreen91 , on the first part of the issue is it okay to use the safe navigator, then we now have `@this_case&.versions` in place of `@this_case.try(:versions)`

rlgreen91:
Eh, I would just stick with the `try` method.  In all honesty, this is actually code that we should fix at a deeper level - we should have some null object that can be returned and still allow this to function correctly without a try or safe navigator.

PR #1562, opened 2018-05-21
Case versions solves #1443

StlMaris123:
In your PR did you:
This pr refactors the history method on cases controller


StlMaris123:
Hi @trystant , @the refactoring suggested fails with undefined method `-`, I have tried to Googled around but so far I haven't seen a use case o how to use it to reverse objects in ruby

rlgreen91:
@StlMaris123 Yeah, so we can't use the `-` operator here since created_at is a DateTime.  My bad.  This should perform the sorting at the database level so we can shave off time and memory:

```
@this_case.try(:versions).order(created_at: desc)
```

Issue #1576, opened 2018-05-25
Treat date as a required case parameter on model level


rlgreen91:
Date (of incident) is a required parameter for a case.  However, there are two places that do not reflect that and should be changed:
* in `app/models/case.rb`, within `case_date_cannot_be_in_the_future`, there is no need to check if the date is present
* add a spec verifying that creating a case without a date will fail

PR #1610, opened 2018-06-11
Add spec to test validate future dates

bmwenda:
#### What does this PR do?
Addresses issue [1576](https://github.com/EBWiki/EBWiki/issues/1576)

#### Notes
- A spec already exists to verify that a case cannot be created without a date. So I added one to verify that a case cannot be created with a date in the future
- I found it necessary to keep the logic that checks if date is present in `case_date_cannot_be_in_the_future`. The reason is that since we are checking that `date > Date.current`, then we should verify that date is not nil. I'm happy to hear other approaches/suggestions  on this though

In your PR did you:

### DIFF
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -31,8 +31,7 @@ class Case < ActiveRecord::Base
   searchkick _all: false, default_fields: ['*']
 
   # Model Validations
-  validates :date, presence: { message: 'Please add a date.' }
-  validate :case_date_cannot_be_in_the_future
+  validate :case_date_validator
   validates :city, presence: { message: 'Please add a city.' }
   validates :state_id, presence: {
     message: 'Please specify the state where this incident occurred before saving.'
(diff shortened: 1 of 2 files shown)


rlgreen91:
Actually, scratch that - I'm not a fan of that approach because to me, it still has the same issue, which is that we're essentially running the date presence validation twice.  I think we'd be better off with a custom date validator that first checks for the date, and returns with an error message if it's not present.  If it is, then it checks that it's not in the future and returns with an error message if it is.  We can namespace it properly as well, something like `app/validators/case_date_validator.rb`.  You can see [here](http://guides.rubyonrails.org/active_record_validations.html#custom-validators) for an example of how to do it.

bmwenda:
@rlgreen91 I made a few changes going by your suggestion. The validator method is only two lines long, so I chose to keep it within `case.rb`. Is that okay with you?

rlgreen91:
@bmwenda Its fine to leave it there.  Can you also add a test case to show that it fails if the date is not present at all?

bmwenda:
@rlgreen91 The spec for nil date already exists. https://github.com/EBWiki/EBWiki/blob/4036ed90358372c30073da949c9751dab7603613/spec/models/case_spec.rb#L7

rlgreen91:
Great, thanks!

Issue #1663, opened 2018-06-25
Write rake task to determine null fields that shouldn't be null


rlgreen91:
Write a rake task that, given a model and a column name, will return if there are any database records that have a null value for the column.  For example, giving the rake task the model `Case` and the column name `summary`, it should return how many `Case` records exist that that currently have a `null` value for `summary`.

PR #1684, opened 2018-06-27
Create null fields counter

bmwenda:
#### What this PR does
Adds a rake task that can count the number of NULL fields for given model's column.

`rake count_null_fields[model,column]`

Addresses issue [1663](https://github.com/EBWiki/EBWiki/issues/1663)

In your PR did you:


lib/tasks/null_fields_counter.rake
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+desc 'Count null fields in a column'
+task :count_null_fields, %i[model column] => :environment do |task, args|
+  abort 'Model or column param missing' if args[:column].blank? || args[:model].blank?
+  model = args[:model]
+  column = args[:column]
+  begin
+    count = model.constantize.where("#{column}": nil).count
```
rlgreen91: Let's change this to `model.constantize.where("#{column}": nil).size`  In general, we want to use `size` over `count` for performance considerations; it may not make much of a difference here, but it's a good habit to get into [(more detail here)](https://work.stevegrossi.com/2015/04/25/how-to-count-with-activerecord/)

rlgreen91:
Please add a spec for the rake task.  If you look at the spec for the `add_blurb` rake task, you should see how we can create cases with missing fields - that should be a good starting point with creating a spec for this task.

bmwenda:
@rlgreen91 Added specs and did some refactoring. I extracted the logic from the rake task to make it reusable and as a [best practice](https://edelpero.svbtle.com/everything-you-always-wanted-to-know-about-writing-good-rake-tasks-but-were-afraid-to-ask)

rlgreen91:
Perfect!  Yeah, I came across that article a few weeks ago, it's got some good stuff.

PR #1693, opened 2018-07-04
sort collection in drop down fixes #1647

### LINKED ISSUE #1647: Sort collections in drop downs
In each form where we render a drop down, the values in the collection should be alphabetized.

This has already been done for the agency drop down in the case form.  It uses the service object `SortAgenciesOrdinally` to ensure that values beginning with a number are handled properly.  You can probably just rename that service object to something like `SortCollectionOrdinally` and then call it each time you need to in the various controllers.

rachaelkiteme:
In your PR did you:


rlgreen91:
Use `SortCollectionOrdinally` for other collections too.  Any time we call `State.all`, `Category.all`, or something similar in a controller.  Probably `Case.all` is the only one I'd leave the way it is.

rlgreen91:
One last thing - in the `SortCollectionOrdinally` service object, change `agencies` to `collection` and `agency` to `element`.

rlgreen91:
I still need you to change `agencies` to `collection` and `agency` to `element` within the code for `SortCollectionOrdinally`.  Additionally, the changes for the redo and undo links shouldn't be present.

PR #1700, opened 2018-07-04
Add Editorconfig

bmwenda:
#### What this PR does
Adds a configuration file to make it easier to maintain EBWiki coding styles. Uses standard rubocop guidelines and can be extended as need be. See [documentation](https://editorconfig.org/) for details

In your PR did you:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# Coding styles for EBWiki
+# editorconfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+
+[*.txt]
+indent_style = tab
+indent_size = 4


rlgreen91:
@bmwenda I'm not so sure that this is necessary.  We do use Rubocop with some custom rules, and the contributor can always run `rubocop -a` before submitting their PR if they like.  Or they can wait for CodeClimate to pick up on the issues.

bmwenda:
@rlgreen91 I think the config can complement rubocop. It helps in automatically setting the rules rubocop checks for in each file. For example, with the editor config, opening a new file with a `.rb` extension will have it set to use two spaces. Also, when you save a ruby file, a new line will be inserted at the end of the file and any trailing white spaces will be removed. 
In as much as you can run `rubocop  -a`, I believe we can reduce some of the errors we need to correct as soon as we can, the editor config would be useful in that sense.

This is just my opinion, I'd be happy to close the PR if you still see this as redundant.

rlgreen91:
Oh, so kind of like the Sublime language-specific preferences.  Interesting.  No, I'll go ahead and merge it.

Issue #1658, opened 2018-06-25
Add suggested title functionality.


rlgreen91:
Note: This issue is 5 of 6 that need to be completed in a particular order for a feature. Each issue should be closed and the corresponding PR merged before the next issue is begun.

On Reddit, users can create posts based on linked content.  When creating the post, the user can click use suggested title which will populate the title field with a suggested value.  It would be great to add that to this.  Write a Javascript function that is triggered on-click that will populate the title field with a suggested value.  It would be easiest to pull the value from the `<title>` tag in the `<head>` element.

claracodes:
Hi I would like to give this a try! Could you specifiy where to add this? To new cases?

> Write a Javascript function that is triggered on-click that will populate the title field with a suggested value. It would be easiest to pull the value from the <title> tag in the <head> element.

PR #1712, opened 2018-07-05
validate url presence solves#1658 

StlMaris123:
In your PR did you:
This Pr validate url presence
- Solves issue  #1658

### DIFF
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -41,6 +41,9 @@ class Case < ActiveRecord::Base
   validates :subjects, presence: {
     message: 'at least one subject is required'
   }
+  validates :links, presence: {
+    message: 'at least one link is required'
+  }
   validates :summary, presence: {
     message: 'Please use the last field at the bottom of this form ' \
       'to summarize your edits to the case.'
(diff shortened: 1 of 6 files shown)


StlMaris123:
Hi @rlgreen91 , Am kindly requesting for more information on this issue on some this that are not clear:

1. Do we have to create a link creation action?
2. Is the link creation form in `new cases form` the one we are using to create the links?
3. At what point will the link be created? Should we have a separate `submit` button for the links on `new cases form`

1.

rlgreen91:
So, the only place we create links is as part of creating or editing cases.  We use Simple Form helper `simple_fields_for` which allows us to create the links as part of creating cases.  See [here](https://github.com/plataformatec/simple_form) and [here](http://api.rubyonrails.org/v5.2.0/classes/ActionView/Helpers/FormHelper.html#method-i-fields_for) for more details.

StlMaris123:
Hi @rlgreen91 , this issue has really helped to understand more on associations....also more on `gem 'simple_form'` and `gem 'cocoon'`

rlgreen91:
That's good to hear!  This looks good, so I'll go ahead and merge.

Issue #1653, opened 2018-06-23
Remove database call from case_helper


rlgreen91:
In `app/helpers/case_helper.rb` we call `Gender.all` to populate the drop-down menu.  That database call shouldn't be there, and really the whole method doesn't need to be in a helper.  Instead, we should handle the drop-down method for gender the same way we do for states and agencies - with a database call in the controller and then passing it along as needed to the view.

PR #1753, opened 2018-07-12
refactor drop down in cases controller solves#1653

StlMaris123:
In your PR did you:
This pr sorts gender using SortCollectionOrdinally.
solves  issue #1653

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -12,6 +12,7 @@ class CasesController < ApplicationController
     @agencies = SortCollectionOrdinally.call(Agency.all)
     @categories = SortCollectionOrdinally.call(Category.all)
     @states = SortCollectionOrdinally.call(State.all)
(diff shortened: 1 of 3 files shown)


StlMaris123:
Hi @rlgreen91 , I was looking at options we have for removing drop_down from the cases helper. I thought of using a concern but they are used to refactor controllers. I researched on ways to refactor views and its either via helpers or partials. Could we move the method to another helper or we just leave it there?

rlgreen91:
Hey @StlMaris123, so rather than do `Gender.all` in the helper, it should be done in the controller.  This is for the form, if I remember correctly - so if you look at the controller action `new`, you'll see we already pass along instance variables for `agencies`.  We should do the same thing for `genders`.  This will require a change to `cases_controller` and the form.

StlMaris123:
Hi @rlgreen91 , did you mean something close to what I have done?

rlgreen91:
Yes, this looks good.

Issue #1655, opened 2018-06-23
Make links polymorphic (3 of 9)


rlgreen91:
Note: This issue is 3 of 9 that need to be completed in a particular order for a feature.  Each issue should be closed and the corresponding PR merged before the next issue is begun.

This issue corresponds to adjusting the schema and model to make `Links` polymorphic.  Create and run a migration with the following methods:

```
def up
  rename_column :links, :case_id, :linkable_id
  add_column :links, :linkable_type, :string
  Links.reset_column_information
  Links.update_all(:linkable_type => "Case")
  add_index :links, [:linkable_id, :linkable_type]
end

def down
  remove_index :links, [:linkable_id, :linkable_type]
  rename_column :links, :linkable_id, :case_id
  remove_column :links, :linkable_type
  add_foreign_key "links", "cases"
end
```

Additionally, update the `Link` model to have the relation `belongs_to :linkable, polymorphic: true` and the `Case` model to have the relation `has_many :links, as: :linkable`.

rlgreen91:
@rachaelkiteme Please note that I have edited the migration text above.

rachaelkiteme:
noted

PR #1755, opened 2018-07-12
add migration for jurisdiction enum[WIP]

### LINKED ISSUE #1344: Add migrations for jurisdiction enum
**Note: This is issue 2 of 7, and the preceding issues in this project must be closed before you can begin this one.**

Run the following migrations to create an ActiveRecord and PostGres enum for jurisdictions:

`bundle exec rails generate migration AddJurisdictionToAgencies jurisdiction:jurisdiction` and

```
class AddJurisdictiontoAgencies < ActiveRecord::Migration
  def up
    execute <<-SQL
      CREATE TYPE jurisdiction AS ENUM ('none', 'local', 'state', 'federal', 'university', 'private');
    SQL

    add_column :agencies, :jurisdiction, :jurisdiction, index: true
  end

  def down
    remove_column :agencies, :jurisdiction

    execute <<-SQL
      DROP TYPE jurisdiction;
    SQL
  end
end
```

rachaelkiteme:
In your PR did you:
 this pr adds migration for jurisdiction enum fixes #1344 


rlgreen91:
@rachaelkiteme I see the commit for the new schema file, but I don't think it was pushed as part of the PR.

rachaelkiteme:
When I merged the the previous versions were disregarded and replaced with the result of the merge. I realized I have to ensure that next time am doing a merge conflict I should make sure I keep the highest version number of the two.

rlgreen91:
Hmm, this is still running into that weird schema error about not being able to drop tables.  Try this set of commands and see if it fixes everything:

```
rake db:drop
rake db:create
rake db:migrate
```

rachaelkiteme:
@rlgreen91 you were to pull this and help me identify the reason why once I run the commands listed above schema.rb file does not update. This will help me fix #1655 too.

rlgreen91:
@rachaelkiteme Try merging master into this branch.  Then, run `rake db:drop` `rake db:create` `rake db:migrate` in that order to regenerate the schema.

rlgreen91:
Ok, so in plain English, here's what happened: the migrations were working fine, the issue is that `db/schema.rb` apparently can't pick up on database-specific things like ENUMs, just what ActiveRecord understands.  So, to be able to view the correct setup, we needed to switch to recording database info in SQL using `structure.sql` rather than in Ruby using `schema.rb`.  So we did that, but we also needed to update config files for Rubocop, Travis, and Heroku to account for the change.  Otherwise everything works the same, just swap out `schema` with `structure` when running rake commands that call for it like `db:structure:load`.

For more info: https://stackoverflow.com/questions/48508905/enumerated-types-with-activerecord-and-postgresql and http://rubyofftherails.blogspot.com/2016/02/good-reasons-to-replace-dbschemarb-with.html and https://github.com/heroku/heroku-buildpack-ruby/issues/543.

rlgreen91:
Oh yeah - I'm leaving the double quotes as is and ignoring CodeClimate because I'm not sure if Heroku will require it to be a double quoted string.

PR #1762, opened 2018-07-13
Automatically redirect url after case title is updated

### LINKED ISSUE #1106: Automatically redirect URLs after case title is updated
To automatically redirect older case URLs after the case title is updated, add the following if statement to the `show` method in the `articles` controller:

```
# If an old id or a numeric id was used to find the record, then
# the request path will not match the article_path, and we should do
# a 301 redirect that uses the current friendly id.
if request.path != article_path(@article)
   return redirect_to @article, :status => :moved_permanently
end
```

You should be able to test this in the controller specs.

rachaelkiteme:
In your PR did you:
FIxes #1106


rlgreen91:
I believe `@case` should be `@this_case` - that should fix Travis. Please add the feature spec when you get a chance.

rlgreen91:
@rachaelkiteme for codeclimate, don't worry about the identical blocks for undo and redo links.  Just make the change to use a guard clause in line 41 for cases controller.

rachaelkiteme:
Am redoing the specs

rlgreen91:
@rachaelkiteme What's the status on this PR?

rachaelkiteme:
Ready for review.

rlgreen91:
Where are the specs?

rlgreen91:
@rachaelkiteme What's the status on this PR?  You said that the specs are done but they aren't included in the files listed above.

rachaelkiteme:
Hi @rlgreen91 , am trying test the redirect given the condition ` request.path != case_path(@this_case)` but the `response.status` is a 200, because its displays the status when the page is rendered, which is a success. Please guide me on how to do the test.

rlgreen91:
@rachaelkiteme I would add an assertion to that test verifying that the title of the page after the redirect is the new title for the case.  At that point it's passed.

rachaelkiteme:
Still not getting it right, cause am getting same output. Maybe you can help with a reference link?

rlgreen91:
You could try testing for the redirect here: https://stackoverflow.com/questions/10955474/how-to-test-for-a-redirect-with-rspec-and-capybara/28986777

Or, you could assert that there is a div containing the updated title present on the page.  Can you push your code for the specs?  I can't really advise you on a spec that I can't see.

rlgreen91:
@rachaelkiteme what's the status on this PR?

rlgreen91:
Hey, so for the feature spec, it's not so much that we need to test IDs work - rather, what we need is a spec that tests that if we access a case using it's old title, that we'll be redirected to the case page.  Does that make sense?

rachaelkiteme:
Yes, sure it does.
Have a question, why is it that when we run `visit "/cases/#{this_case.id}"` on the console it returns `nil`?

rlgreen91:
It's most likely because you're no longer in the test environment.

rlgreen91:
@rachaelkiteme What's going on with this PR?  Do you need help with anything?

rachaelkiteme:
Yes, I do will be very grateful, was looking at it yesterday but the output
is still disappointing, Did I add the changes in the right way?
Regards
Rachael

rlgreen91:
Ok, so here's some pseudocode:

```
create a case and navigate to the page
save title of case as "old title"
update case title and save new title as "new title"
visit '/cases/old_title'
verify assertion that you are redirected to page with url '/cases/new_title'
```

To verify the assertion, you can use something like `expect(current_path) to equal 'cases/new_title'`

rlgreen91:
Ah, let's change that last line to be `expect(page).to have_current_path('/cases/new-title')` - using that matcher will trigger Capybara's waiting behavior, which should be long enough for the redirect

rlgreen91:
Great!  One last thing - can you merge master into this branch?  I suspect there's a couple of things that have changed, particularly with that `before_filter`

rlgreen91:
Okay, thanks for adding master in.  One thing to note - we still want to make sure that we're not defining instance variables in private methods.  Update `find_case` to work the way `load_commentable` does in the comments controller, and remove the `before_filter`.

rlgreen91:
Why are you returning in the `show` action?  That's not going to execute the rest of the code at all.

rlgreen91:
@rachaelkiteme What's the status on this PR?

rachaelkiteme:
Am getting this error and I know how to go about it 
```Bullet::Notification::UnoptimizedQueryError:
       user: rirrie
       GET /cases/old-title
       AVOID eager loading detected
         Case => [:comments, :subjects]
         Remove from your finder: :includes => [:comments, :subjects]
       Call stack
         /home/rirrie/EBWiki/spec/features/user_visits_case_page_spec.rb:9:in `block (2 levels) in <top (required)>```

rlgreen91:
If you run into an error, please push the code to the PR.  Otherwise it's difficult to see what exactly the issue is.  At best guess, I would say you could try using a `where` or `find_by` clause would allow you to still do the eager loading.  Searching for how to use `includes` with friendly_id should turn up some good results.

trystant:
I'm going to close this PR out and re-apply the lessons learned in a new branch.

Issue #1622, opened 2018-06-16
Remove conditional from agency show page


rlgreen91:
in `app/views/agencies/show.html.erb`, we have a conditional that checks to see if the agency jurisdiction is present and if so, displays it.  We can remove the conditional surrounding the display since the helper we use for display does that check for us.  We can remove the conditional for the website as well.

PR #1803, opened 2018-07-18
refactor agency show page solves #1622

StlMaris123:
In your PR did you:
This pr removes the conditional check on  if the jurisdiction is present since there is a similar method in the helper.
- Solves issue #1622

### DIFF
--- a/app/views/agencies/show.html.erb
+++ b/app/views/agencies/show.html.erb
@@ -25,12 +25,13 @@
 </p>
 
 <p>
-  <%= show_non_blank_fields "Notes", @agency.description %>
+  <%= show_non_blank_fields "Website", @agency.website %>
 </p>
 
 <p>
-  <%= show_non_blank_fields "Website", @agency.website %>
+  <%= show_non_blank_fields "Notes", @agency.description %>
 </p>
+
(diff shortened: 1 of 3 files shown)


app/views/agencies/show.html.erb
```diff
@@ -27,12 +24,6 @@
   <%= show_non_blank_fields "Email", @agency.email %>
 </p>
 
-<p>
-  <% if @agency.website.present? %>
-     Website: <%= link_to @agency.website, @agency.website %>
```
rlgreen91: @StlMaris123 You haven't actually made the change to the helper, just deleted this section.

StlMaris123:
Hey @rlgreen91 the `show_non_blank ` is already called on the agency show view

app/views/agencies/show.html.erb
rlgreen91: @StlMaris123 No, what I'm saying is that in line 13, you use the helper to display the jurisdiction if it's present.  I need you to do the same thing for the website - use the helper to display the website if it's present.  Replace lines 30-35 with that.

rlgreen91:
@StlMaris123 Can you put it back in the order it was in before?  With website before notes?

Issue #1674, opened 2018-06-26
Create service object to process new users (1 of 3)


rlgreen91:
Note: This issue is 1 of 3 for the given topic, and must be followed in the order listed.  The preceding issues must be completed and merged before continuing to the next issue.

Create a service object that replicates the functionality in `UserObserver`.  That is, given a user object, the service object should add the user to Mailchimp and send the welcome email.

rlgreen91:
This is the article I read when I was introduced to the concept a couple of years ago - https://brewhouse.io/blog/2014/04/30/gourmet-service-objects.html.  It holds up pretty well I think.  There's really just a few things you want your service objects to do: 
* follow the Single Responsibility Principle (even if that responsibility is just coordinating other service objects); 
* be as functional as possible, in that you provide objects as inputs and do some operation without altering the objects - if you're altering the objects, then you're essentially dealing with business logic*
* have a single point of entry - if you look in our services folder, we have a Service concern that basically maps a class method `call` with the supplied arguments to an instance method that initializes the service object and then invokes an instance method `call` with the supplied arguments.  Each of the service objects we have now includes `Service` and defines an instance method `call`, and yours should too.

If you want to look into this more, try searching for "operations," "commands," and "interactors" since those are other terms for this type of design pattern.

*Yeah, I probably shouldn't say business logic since that's a term applied to both models and service objects.  Really, what it comes down to is the idea that a model represents your data, and whether/how it responds to CRUD operations in the database, while a service object represents what your application does with the data.

StlMaris123:
First thing on my list tomorrow!!

PR #1917, opened 2018-07-29
Begin implementation for Services #1674 

StlMaris123:
In your PR did you:
Re-implement UserObserver as a service object


StlMaris123:
Hello @rlgreen91 , Am kindly requesting for your help moving forward with this. Am still struggling with services

rlgreen91:
@StlMaris123 Ok, so if you look within app, you should see a folder titled `services`.  With the exception of `service.rb`, each of the files defines a service object - `AddUserToMailchimp`, `DetermineVisitsToCases`, and `SortCollectionOrdinally`.  What you need to do is create a service object like that called, let's say `OnboardUser`.  That object, given a user as the parameter, would be responsible for doing what's outlined in `UserObserver` now - adding them to Mailchimp and sending the email.  Why don't you try making that first - the new file you created is already most of the way there.  You just need to make it look like the other ones in the folder.

StlMaris123:
Hi @rlgreen91 , I have made some changes. How do we call the service? I tried as a callback but it does not work

rlgreen91:
Ok, a couple of things - it's probably failing because it's not finding the file.  It should be `app/services/onboard_user.rb`, not `app/services/on_board_user.rb`.  The name of the module you're trying to load doesn't match the file name convention Rails is expecting.

The second thing is that this shouldn't be a module - it should be a class.  Lastly, we actually won't call this in the model - instead, we'll call this from the controller.  So what you'll do is call this in the registrations controller.

rlgreen91:
Ok.  Be sure to call the service object from the registrations controller, rather than in the model.  Additionally, also change the name of the file.  Finally, rather than commenting the line in `config/application.rb`, you can just remove it.

StlMaris123:
Hi @rlgreen91 , only issue I have in putting it in the registrable controller is getting the user object

rlgreen91:
@StlMaris123 You can still access it there, it's just that Devise calls it a resource in the controller rather an a user.

Issue #1732, opened 2018-07-07
Update Devise gem to 4.2.1


rlgreen91:
Note: This issue is 5 of x for the given topic, and must be followed in the order listed. The preceding issues must be completed and merged before continuing to the next issue.

Update the Devise gem to 4.2.1, specifying that version in the Gemfile.

Within `rails_helper`, change `Devise::TestHelpers` to `Devise::Test::ControllerHelpers, type: :controller` and add `Devise::Test::IntegrationHelpers, type: :feature` within that same block.

PR #1999, opened 2018-08-07
Update Devise gem to 4.2.1 #1732

StlMaris123:
In your PR did you:
Update Devise gem to 4.2.1
- Solves #1732


StlMaris123:
Hi @rlgreen91 , learnt that in rspec 3, if a test is actually passing but has been marking as pending it will be classified as  failed. Thats why I removed the pending spec

rlgreen91:
@StlMaris123 thanks for the heads up.  Seems silly but not something I can change.

Issue #1977, opened 2018-08-06
Update paper_trail gem to 5.2.3 (2 of 3)


rlgreen91:
Note: This issue is 2 of 3 for the given topic, and must be followed in the order listed. The preceding issues must be completed and merged before continuing to the next issue.

Update paper_trail to 5.2.3.  In `spec/mailers/user_notifier_spec.rb`, change `PaperTrail.enabled = false` to `Case.paper_trail.disable` and `versions.last.whodunnit` to `versions.last.paper_trail.whodunnit`.*  In `app/views/user_notifier/send_followers_email.html.erb`, change `@this_case.versions.last.whodunnit` to `@this_case.versions.last.paper_trail.whodunnit`.  In `app/views/cases/history.html.erb`, change `version.whodunnit` to `version.paper_trail.whodunnit`.

*This spec is part of a PR at the time this issue was written, so the lines noted here may be gone.

PR #2001, opened 2018-08-07
Update paper_trail gem to 5.2.3 #1977

StlMaris123:
In your PR did you:
Update paper_trail gem to 5.2.3 
Solves #1977


StlMaris123:
Hi @rlgreen91 , I have reverted the changes you had requested, they were having the failure ` #<Double (anonymous)> received unexpected message :whodunnit with (no args)
`

rlgreen91:
@StlMaris123 you still need to make the changes listed in the issue, otherwise the app will break on this going forward.  If it causes a test to fail put the result here or let it fail in CI.  Otherwise, I can't provide suggestions on how to fix it.

StlMaris123:
I have struggled with this issue, so I put as pending. I tried googling but could not find a solution: 
```ActiveRecord::StatementInvalid:
       PG::InFailedSqlTransaction: ERROR:  current transaction is aborted, commands ignored until end of transaction block
       : SAVEPOINT active_record_2
```

rlgreen91:
Agh, I actually came across this maybe a couple of weeks ago?  The error is kinda misleading, but basically if PostGres can't execute a SQL statement due to invalidity, then it just automatically aborts all future transactions that occur in the same connection.  When exactly where you encountering this?  If you were running the server locally, or the console, or debugging during a test, quitting and restarting should be enough to stop it so you can fix the SQL.  If not, let me know and we can look some more.

StlMaris123:
Hi @rlgreen91 , the error occurs when running the tests. It is the only spec failing

rlgreen91:
Ok.  There's some other stuff going on with history pages in general, so we'll just leave this as pending and merge.

Issue #1604, opened 2018-06-11
Add official colors as CSS variables


rlgreen91:
Add the corresponding colors as CSS variables in the top-level stylesheet:
* F65732: orange
* 666666: gray
* 000000: black
* 4D4D4D: dark gray

PR #2015, opened 2018-08-08
Add field title in case form links section

### LINKED ISSUE #1657: Update case form to include field for title in links section (4 of 9)
Note: This issue is 4 of 9 that need to be completed in a particular order for a feature. Each issue should be closed and the corresponding PR merged before the next issue is begun.

Update the case form and controller to include a field for titles in the links section.

rachaelkiteme:
In your PR did you:
Fixes #1657 
This pr updates case form with title field in links section


rachaelkiteme:
case form links section
title field

rlgreen91:
Can you put the URL field above the title field?  We'll be adding some functionality in another issue that makes it better to have it in that order.  And yes, the title field shouldn't be required as of now.

While we're here, can you make sure the orange used for the "Add a Link" button is the same as the one we used elsewhere, that we have a variable for?

rlgreen91:
Sorry, is it the same orange @rachaelkiteme?

rachaelkiteme:
Yes sure, can confirm here #1604

Regards
Rachael

rlgreen91:
Oh, ok, so rather than copy the hex value, just use the variable name we gave it.  `orange` is assigned to that hex number in the stylesheet, so you should be able to just the value to `orange` and confirm that it's the correct color.  Although we didn't set the value of `white`, that is a variable that is automatically added by the processor so you can use that instead of the hex value as well.

rlgreen91:
Why are you modifying `bootstrap-mods.scss`?

rachaelkiteme:
Oh, thats what caused the error!! The reason I changed `bootstrap-mods.scss` is that I thought thats where we do bootstrap modifications.

rlgreen91:
No worries.  No, we currently kinda make the changes in the corresponding stylesheet for the model.  Eventually we'll get around to organizing those a bit more clearly.

PR #2079, opened 2018-08-16
add gem pundit 2.0 having fixed gemfile.lock conflict

turuthivic:
In your PR did you:
The original pull request was to add gem pundit 2.0 but it came with conflict hitches(next time I will look out for the "Able to Merge notification during a pull request)
After scratching my head for sometime the Gemfile.lock conflicts are resolved


rlgreen91:
@turuthivic It's generally preferred that if you have to make changes to your PR, that you do so within the original PR rather than creating a second one.  All you need to do is push your new or updated commits to the same branch on the remote repo and it'll update the PR.  I suggest doing that here, by pulling down master, merging it into the other branch you have open, and then pushing those changes up here.  The guide [here](https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging) should be useful.

Issue #1345, opened 2018-03-08
Set default for jurisdiction enum (3 of 7)


rlgreen91:
**Note: This is issue 3 of 7, and the preceding issues in this project must be closed before you can begin this one.**

Set the default value for the jurisdiction enum in the `agency` model to `'none'` - this should be done via a migration.

rlgreen91:
This address part of the change discussed in #1332

rlgreen91:
@rachaelkiteme What's the status on this issue?

PR #2081, opened 2018-08-16
set default value for jurisdiction enum

rachaelkiteme:
In your PR did you:
Fixes: #1345 
Run a migration to set the default value for jurisdiction enum to "none"


rlgreen91:
@rachaelkiteme The default value for the value for `jurisdiction` should be `none`, not `jurisdiction_type`.  Just change the column name and run `rake db:migrate:redo` to fix the migration instead of creating a new one.

rachaelkiteme:
But its a irreversible migration.
Help me understand why we use `jurisdiction` while else we changed it to
`jurisdiction_type` some time back.

Regards
Rachael

rlgreen91:
Is it?  Then try just adding a line to change the default value of the `jursidiction` column to `'none'`.  

Here's what's going on.  We want to use what are called enum types to represent the possible jursidictions an agency might have (later we'll do this with subjects and gender).  Right now, we use a gem to provide that functionality.  However, ActiveRecord has enum types built-in, so it doesn't really make sense for us to use a gem to do that.  We want to switch, which will take a few steps, but at the same time we want to ensure that there is 0 downtime for this feature - that people can still assign a jurisdiction to an agency.  That's not too bad - this whole set of issues is written to allow us to make that change with 0 downtime.  However, there's already a value set for the name `jurisdiction` - if we want the final version, with the ActiveRecord enums, to have that name, then we need to rename the original so we don't end up with conflicts.  That's where it can get confusing.  Here, we're just declaring a default value of `none` for the new ActiveRecord jurisdiction enum before we start using it.

If you get a chance, try reading all of the issues in this set and go through them in order.  Hopefully it'll become clear how they connect to one another.

Issue #1893, opened 2018-07-26
Create a versions controller with revert action (2 of 5)


rlgreen91:
Note: This issue is 2 of 5 for the given topic, and must be followed in the order listed. The preceding issues must be completed and merged before continuing to the next issue.

Create a new versions controller with a revert action.  Let's do the following:
* start by copying over the undo action in the cases controller - in fact, you can delete that
* change `@case_version` to `@version`
* change the flash success message to `Reverted changes`
* Under the flash success message, add the following code: `flash[:reversion] = @version.next
* Replace the ensure block with `redirect_to :back`

Then we can update the routes file with this code: `post 'versions/:id/revert' => 'versions#revert', :as => 'revert'.  Just specifying the URL format, the associated controller action, and the path name.

PR #2090, opened 2018-08-17
move undo action in cases controller to versions controller solves #1893

StlMaris123:
In your PR did you:
move undo action in cases controller to versions controller
-solves #1893


rlgreen91:
@StlMaris123 For codeclimate, you can disable the check for too many lines in a method.  However, please add top level documentation as requested.

rlgreen91:
Hmm...so we actually need to account for the fact that in the future we'll be enabling PaperTrail for other models.

Let's nest this under cases in the routes, and move this controller inside of a module `Cases`.  In the routes, rename `:id` to `:case_id` so that it's clear which model this route is nested under.

StlMaris123:
Hi @rlgreen91 , Is this what you were talking about?

rlgreen91:
Yes, the only thing left to do is put the controller in a module `Cases`

rlgreen91:
Perfect

Issue #1237, opened 2018-01-05
Enable friendly-id for states


rlgreen91:
We currently use the gem friendly-id to create friendly slugs for cases, and search by them as well.  Add this functionality to states.  
This will involve updating the model to include friendly-id slugs and the controllers and service objects to search by slug rather than ID.  We shouldn't be searching for states within the views themselves but if for some reason we are, add a comment noting where.

To see how to configure friendly_id use the guides: https://github.com/norman/friendly_id#usage.  It's already been installed, no need to do that part again.

PR #2123, opened 2018-08-21
enable friendly ids for states #1237 

StlMaris123:
In your PR did you:
enable friendly ids for states
- Solves #1237


rlgreen91:
That's fine to use name.  But yeah, there's plenty of places where we do call `State.find` or something like that.  That's what you should be replacing - anytime we find an instance of `State` using the ID.

StlMaris123:
Hi @rlgreen91 , I did a global search, there is nowhere we are doing `State.find`

rlgreen91:
Ah, also, your PR failed in CI because we need to add a `slug` column to states, despite what the docs imply.

rlgreen91:
@StlMaris123 did you run the migration?  There should have been a change to `db/structure.sql` as well.

PR #2298, opened 2018-08-30
set instance variable in cases controller action

### LINKED ISSUE #2157: Set instance variables in cases controller actions
In cases controller, rather than have a `before_action` that defines `@this_case` for certain methods, define that variable directly in each of the corresponding actions.  The actual code used to find the case can be hidden via a private method, it's just that `@this_case` needs to be defined directly in the action.  In that case, the private method would just return the value rather than setting `@this_case`.  

Additionally, when searching for the case, it should be done using `friendly.find`, not `friendly.find_by_id`.

rachaelkiteme:
In your PR did you:
Fixes #2157 

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -3,7 +3,6 @@
 # Cases controller. Containing really complex index method that needs some
 # Refactoring love.
 class CasesController < ApplicationController
-  before_action :find_case, only: %i[show edit update destroy history]
   before_action :authenticate_user!, except: %i[index show history followers]
 
   def new
@@ -84,6 +83,7 @@ class CasesController < ApplicationController
   end
 
   def destroy
+    @this_case = Case.friendly.find_by_id(params[:id])
     if @this_case
       @this_case.destroy
       flash[:success] = 'Case was removed!' # {make_undo_link}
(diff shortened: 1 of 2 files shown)


rlgreen91:
Thanks for catching the typo.  So, for the `destroy` and `history` actions, we should be using `Case.friendly.find` since we're really using the slugs, not the IDs.  Check that you have the latest version of the routes file - we've updated some of the routes to use `case_slug`, so it's clear what parameter we're using.

Now, if you're going to query the database for the case directly in the action, then there's no need to have the method `find_case`, since that's doing the same thing.  Or, you can keep the private method, and just set it equal to the instance variable in the actions.  Either one is fine, it's your call.

app/controllers/cases_controller.rb
```diff
@@ -84,23 +83,28 @@ def update
...
 
   def destroy
-    if @this_case
-      @this_case.destroy
-      flash[:success] = 'Case was removed!' # {make_undo_link}
-      UserNotifier.send_deletion_email(@this_case.followers, @this_case).deliver_now
-    else
+    begin
+      @this_case = Case.friendly.find(params[:id])
+      if @this_case
```
rlgreen91: You don't need this if statement if you rescue the not found error.

Issue #2148, opened 2018-08-23
Create a null case object


rlgreen91:
Create a new directory called app/null_objects and a new file `app/null_objects/case.rb`.  Create a new class, `NullObjects::Case` that has the following initializer:

```
def initialize
  Case.none
end
```

Then, we can update the `history` action of the cases controller as such:
- replace `try(:versons)` with a private method that will check for the presence of versions on a case; if there are none, return a new instance of `NullObjects::Case`, otherwise, return the versions of the case as the variable
- remove the check in the unless statement for case versions

In the end, the first statement should look something like this: `this_case.private_method.order(created_at: :desc)`

k-stewart:
I see that StlMaris123 made some progress on this.  What remains to be done?

StlMaris123:
Hey @k-stewart , i closed my pr, it is okay to start from scratch

k-stewart:
Thanks for the update!

k-stewart:
Hi @rlgreen91, I'm running into a couple problems with this ticket!

The most pressing one is that, currently, `agency.nearby_cases` doesn't return cases. It returns agencies near to the agency in question. What should `nearby_cases` return?

Should it return cases belonging to any agency that are near the agency in question?
Should it return only cases that belong to the agency?
Should it return only cases without an agency?

rlgreen91:
Hi @k-stewart, sorry for the delayed response!  If you're still wanting to work on this, I'd say that `agency.nearby_cases` doesn't really make sense - do you remember where you saw that in the code?  If not, that's fine, please just let me know either way.

PR #2304, opened 2018-08-30
Create a null case object #2148

StlMaris123:
In your PR did you:
Create a null case object 
- Solves #2148


StlMaris123:
Hi @rlgreen91 I did a few tweaks in this pr:
1. I created the null object in a folder called `null`, the `null_objects` had an issue Uninitilaized constant error.
2. I called the method in the history action instead of calling it on the case object, because that mean i define it in the case model.
What do you think of the approach

rlgreen91:
Hey, so what was the error about Uninitialized constant?  You shouldn't have gotten something like that.  If you had a file `app/null_objects/case.rb`, then this should have worked:

```
module NullObjects
  class Case
    ...
  end
end
```

Also, while your thoughts on `check_case_presence` are good, we don't want to do quite that because we do a redirect (or should be) if the case doesn't exist.  So for right now we should just replace the versions check.

rlgreen91:
@StlMaris123 Please merge master into this branch.  Additionally, there should now be a `null_objects` directory.  Please be sure to add a spec as well.

StlMaris123:
Hi @rlgreen91 , I found out how we were getting the `uninitialized constant NullObjects::Case`, as explained here https://github.com/rails/rails/issues/14382, the nuul_objects is treated as root_path and therefore cannot be used for name spacing.

rlgreen91:
That particular issue isn't relevant here, given that the issue there was not following naming conventions.  Not to mention that we've already added a null_objects directory. :)

I think you're running into issues here because we're trying to replace too much with null object.  For right now, if we find that `this_case` is empty, then we should set `case_history` to `nil`.  Then, when it comes time to see if the case has any versions, that's when we either return the case or the null case object based on the result.  Hopefully that should fix the error.

StlMaris123:
Hi @rlgreen91 , before we were looping through case_history only if a case has versions. But now if a case has no versions or its empty. We are returning Case.none.
The tests are failing because we are trying to loop through Case.none

rlgreen91:
I see.  Hmm, in that case let's trying adding a `versions` method to the null case object that returns `Case.none`

rlgreen91:
Please merge master into this branch.

StlMaris123:
Hi @rlgreen91 , issue here is that 
`Case.none` gives `#<ActiveRecord::Relation []>` , so trying to called versions on its instance will  raise```NoMethodError (undefined method `versions' for #<ActiveRecord::Relation []>)```

StlMaris123:
Hi @rlgreen91 , what is the way forward with this?

rlgreen91:
For right now, let's change the code in the view to check `if @case_history.blank?`.  Once we add presenters we can clean that logic up a bit.

Issue #2115, opened 2018-08-21
Change color of buttons on case show view


rlgreen91:
On the case show view, use the standard orange color for each button to undertake an action (e.g., follow, edit, history).  Feel free to add a black border for the buttons on the left to help distinguish when they begin and end.  For buttons to "undo" an action (e.g., unfollow), use the standard gray color.

Additionally, fix the spacing for the text on "unfollow".

PR #2347, opened 2018-09-03
#2115 updated the button colours on case show view

danuluma:
In your PR did you:

### DIFF
--- a/app/assets/stylesheets/bootstrap-mods.scss
+++ b/app/assets/stylesheets/bootstrap-mods.scss
@@ -1,3 +1,5 @@
+$color-orange: #F65732;
+
 .btn {
     padding: 14px 24px;
     border: 0 none;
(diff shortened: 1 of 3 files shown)


rlgreen91:
@danuluma the change to the Gemfile should not be here.  Additionally, why did you create a new class for the button?  Was it not possible to style it using existing classes?

In the future, if you are making changes to the UI please include a screenshot.

danuluma:
The faker gem was still in the test group and rails server couldn't start if it wasn't in the development environment, hence I moved it there. There's a pr for that but it had not yet been merged into master.

danuluma:
The classes already available are default bootstrap classes, I'm not sure if they have been used in other views and since the changes were intended for the cases view only, I thought it better to create a new class and define its styles in the cases.scss

danuluma:
screenshot from 2018-09-04 04-58-01
screenshot from 2018-09-04 04-57-53

danuluma:
Sorry, forgot to include the screenshots in the pr. I used the $color-orange variable already defined, but it appears to me more of red than orange

rlgreen91:
@danuluma that PR has been merged into master, so please update this branch.  With regard to the buttons, we do apply this styling on other pages, so I would check how they handle it and replicate/share styles as much as possible.  I'm not sure what's up with your colors though - it's definitely more orange when I see that color, and that's true for multiple computers.

PR #2358, opened 2018-09-04
Set instance variable in agency controller actions

### LINKED ISSUE #2158: Set instance variable in agency controller actions
In agencies controller, rather than have a `before_action` that defines @agency for certain methods, define that variable directly in each of the corresponding actions. The actual code used to find the agency can be hidden via a private method, it's just that `@agency` needs to be defined directly in the action. In that case, the private method would just return the value rather than setting `@agency`.

rachaelkiteme:
In your PR did you:
This pr define @agency variable directly in each actions rather than having before_action
Fixes #2158 

### DIFF
--- a/app/controllers/agencies_controller.rb
+++ b/app/controllers/agencies_controller.rb
@@ -2,7 +2,6 @@
 
 # Agencies Controller
 class AgenciesController < ApplicationController
-  before_action :set_agency, only: %i[show edit update destroy]
   before_action :authenticate_user!, except: %i[index show]
   before_action :save_my_previous_url, only: %i[new show edit]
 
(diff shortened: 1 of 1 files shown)


rachaelkiteme:
Question. Why are we defining the variables directly in the actions?

rlgreen91:
In general, it's considered bad practice to hide how you're generating your instance variables (and to use instance variables in partials, but that's another issue we'll tackle later).  Yes you typically want to make your code DRY, but given how often you use and reuse instance variables having a bunch of them defined in private methods is confusing, especially if multiple actions use the same instance variables.  I'd read [this article](https://blog.codeminer42.com/controller-best-practices-dont-hide-instance-variables-5e8bf067156) for a more detailed explanation.

PR #2359, opened 2018-09-04
set instance variable in comments controller actions

### LINKED ISSUE #2159: Set instance variable in comments controller actions
In comments controller, rather than have a `before_action` that defines @commentable for certain methods, define that variable directly in each of the corresponding actions. The actual code used to find the comment model can be hidden via a private method, it's just that `@commentable` needs to be defined directly in the action. In that case, the private method would just return the value rather than setting `@commentable`.

rachaelkiteme:
In your PR did you:
Fixes #2159 

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -2,18 +2,23 @@
 
 # Case comments controller
 class CommentsController < ApplicationController
-  before_action :load_commentable
   before_action :authenticate_user!, only: [:create]
 
   def index
+    resource, id = request.path.split('/')[1, 2]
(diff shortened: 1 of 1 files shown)


rachaelkiteme:
That throws error `undefined local variable or method `resource' for #<CommentsController:0x0000000006675308>`

rlgreen91:
No, what I mean is more like this:

```
def set_commentable
  resource, id = 'whatever this method is'
  whatever the second method is
end
``

Then, in the controller actions, just call `@commentable = set_commentable`.  We set the instance variable in the controller action but still keep the private method.

Issue #1981, opened 2018-08-06
Add optional attribute name parameter to sorting service object


rlgreen91:
Update the service object `SortCollectionOrdinally` to take an optional parameter `column_name`.  This parameter should default to `name`, and whatever value is provided should be used when sorting the collection.  For example, if the value is `name`, then you should sort the collection ordinally based on the name attribute of the corresponding model.  Note that you will also have to pass along the collection as a keyword argument as well - just name it `collection`.

You can test this using the Gender model, which instead of a name has a sex.  

Then, in the controllers, update the lines using `SortCollectionOrdinally` to pass the parameters along properly.  Remember that when using the service object for genders, you'll need to set the `column_name` parameter as sex.

trystant:
The `SortCollectionOrdinally` was passed a `Gender` model, so I have a patch that creates an alias column `:name` so that Gender works in #2028

StlMaris123:
Hi @trystant , should this issue be closed?

rlgreen91:
No, go ahead and make the changes.  We might need this for other models in the future, and adding a bunch of aliases isn't practical.

PR #2373, opened 2018-09-05
Add optional attribute name parameter to sorting service object #1981

StlMaris123:
In your PR did you:
Add optional attribute name parameter to sorting service object 
- Solves #1981

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -48,7 +48,7 @@ class ApplicationController < ActionController::Base
   end
 
   def state_objects
-    @state_objects ||= SortCollectionOrdinally.call(collection: State.all)
+    @state_objects ||= SortCollectionOrdinally.call(collection: State.all, column_name: e.send(:column_name))
   end
 
   def storable_location?
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -48,7 +48,7 @@ class ApplicationController < ActionController::Base
   end
 
   def state_objects
-    @state_objects ||= SortCollectionOrdinally.call(collection: State.all, column_name: e.send(:column_name))
+    @state_objects ||= SortCollectionOrdinally.call(collection: State.all)
   end
 
   def storable_location?
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -48,7 +48,7 @@ class ApplicationController < ActionController::Base
   end
 
   def state_objects
-    @state_objects ||= SortCollectionOrdinally.call(State.all)
+    @state_objects ||= SortCollectionOrdinally.call(collection: State.all)
   end
 
   def storable_location?
--- a/app/services/sort_collection_ordinally.rb
+++ b/app/services/sort_collection_ordinally.rb
@@ -5,7 +5,7 @@
 class SortCollectionOrdinally
   include Service
 
-  def call(collection:, column_name: 'name')
+  def call(collection:, column_name: column_name)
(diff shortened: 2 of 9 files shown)


app/services/sort_collection_ordinally.rb
```diff
@@ -5,7 +5,7 @@
 class SortCollectionOrdinally
   include Service
 
-  def call(collection)
+  def call(collection:, column_name: 'name')
     first_sort = collection.sort_by { |e| ActiveSupport::Inflector.transliterate(e.name.downcase) }
```
rlgreen91: Whoops - what we need to do here is take the string passed as the column_name and use `send` to get the value.  So something like `e.send(:column_name) throughout.

rlgreen91:
Also, let's go ahead and remove the alias from the gender model.  Doing that should confirm that this all works together.

app/services/sort_collection_ordinally.rb
StlMaris123: Hi @rlgreen91 , its not clear to me what you meant

rlgreen91: So, right now we keep calling `e.name` right?  The issue is that we don't know what the column name is going to be.  We need to basically turn it into a method that is applied to an object.  One way you can do that is with the `send` method - pass it a value and it will apply it as a method to whatever is calling it.  So this basically says take the variable `e` and apply the value `:column_name` as a method.  

Hmm, you might need to assign column name variable first, something like `:column_name = column_name`

rlgreen91:
Please merge master into this branch.

StlMaris123:
Hi @rlgreen91 , I have done a change to one of the methods where `SortCollectionOrdinally` has been called to check if that is what you meant

app/controllers/application_controller.rb
```diff
@@ -48,7 +48,7 @@ def log_invalid_token_attempt
   end
 
   def state_objects
-    @state_objects ||= SortCollectionOrdinally.call(State.all)
+    @state_objects ||= SortCollectionOrdinally.call(collection: State.all, column_name: e.send(:column_name))
```
rlgreen91: We don't need the second parameter here, since states have names in our schema.

rlgreen91: Oh, ok - so when I talk about using `send`, I mean in the actual service object code, not here.

rlgreen91:
Ok, so you still need to make the changes in `app/services/sort_collection_ordinally`

app/services/sort_collection_ordinally.rb
```diff
@@ -1,32 +1,33 @@
...
+# frozen_string_literal: true
+
+# Given a relation containing all of the element objects, sort them by name while accounting for
+# ordinal numbers
+class SortCollectionOrdinally
+  include Service
+
+  def call(collection:, column_name: 'name')
+    first_sort = collection.sort_by { |e| ActiveSupport::Inflector.transliterate(e.send(column_name).downcase) }
+    names_with_numbers = first_sort.select { |element| element.name.match(/(\d+)/) }
```
rlgreen91: You sitll need to change this to `send(column_name)` here as well.

app/services/sort_collection_ordinally.rb
```diff
@@ -1,32 +1,33 @@
...
+    sorted_names_with_numbers + names_with_no_numbers
+  end
+
+  private
+
+  # Performs ordinal sort removes number portion from string, then sorts names based on
+  # number then following words
+  def ordinal_sort(collection)
+    name_object_hash = {}
+    collection.each { |element| name_object_hash[element.name] = element }
```
rlgreen91: And here

Issue #2371, opened 2018-09-05
Use proper status codes in create actions


rlgreen91:
In each of our controllers, render `status: :created` when the `create` action is successful.

rlgreen91:
We cannot make this change as specifying the status code requires us to use `render`, which doesn't work because the corresponding action will not run before the page is loaded.  This in turn causes errors when expected variables are missing.

PR #2431, opened 2018-09-12
Use proper status codes in create actions #2371

StlMaris123:
In your PR did you:
Use proper status codes in create actions 
- Solves #2371

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ bower.json
 
 # Ignore DS_Store files
 .DS_Store
+
+# remove database.yml from files being pushed
+config/database.yml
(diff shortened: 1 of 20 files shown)


StlMaris123:
Hi @rlgreen91 , I did try to update the status. However the redirects are changing into status 201 that's why the cases controller is failing.  The rest of the tests are passing because they that particular part has not been tested.

rlgreen91:
Ack, I'll need to close this: just realized that if we only call `render`, then the action associated with it won't be called, meaning we'll be missing variables that it expects to be defined.  I don't know that the tests catch it but if you navigate through your browser manually it should complain.

Issue #2406, opened 2018-09-08
Create a rake task to set the values of comments in versions table


rlgreen91:
Create a rake task that will search the versions table, grab all records for which the `comment` column is `nil`, and set the value to an empty string.  There's a rake task to add a blurb field to cases that you can use as a model of what to do here.

asarbu:
I can work on this :)

yifeiwu:
Why not make this a migration to set the db default? That way you won't need to rerun the rake task.

rlgreen91:
@yifeiwu sorry, just now seeing your comment.  The issue is that because the table already exists, setting the default will only be applied to rows added after the migration is run, not before.  So we still need a way to update the existing records with the default.

PR #2477, opened 2018-10-03
Create a rake task to set the values of comments in versions table

asarbu:
Issue: https://github.com/EBWiki/EBWiki/issues/2406

Create a rake task to set the values of comments in versions table.
The task can be run using: `rake versions:set_nil_comment`

In your PR did you:

### DIFF
--- a/lib/tasks/versions.rake
+++ b/lib/tasks/versions.rake
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+namespace :versions do
+  desc 'Set all nil comment attributes on versions to empty string'
+  task set_nil_comment: :environment do
+    sql_query = <<-END_SQL
+      UPDATE versions
+      SET comment = ''
+      WHERE comment IS NULL;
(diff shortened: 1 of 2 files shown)


rlgreen91:
@asarbu Please add a spec for the rake task.

asarbu:
@rlgreen91 added spec

rlgreen91:
Thanks for adding the spec.   While I'm fine with using straight SQL in the rake task itself, can you use ActiveRecord in the spec, closer to what is in the other specs for rake tasks?  Right now it's not necessarily clear how the spec works if you're not used to using SQL as part of `connection.execute`.

asarbu:
@rlgreen91 I would have used ActiveRecord, but it appears that we don't have a model set for `Version`. Unless that model is added we cannot test this otherwise.

rlgreen91:
Whoops, this one got stuck at the bottom of the pile, but still looks good overall.  Will merge.

Issue #2263, opened 2018-08-28
Update case when link is updated


rlgreen91:
Let's update the parent case when a link is created, updated, or destroyed.  To do that, add `touch: true` at the end of the `belongs_to` association in the link model.  Then, update the copy on the case history view to state that a new version is created any time a link associated with the case is.  Finally, write a feature spec that confirms that updating a link will update the version of a case.

PR #2491, opened 2018-10-11
Update case when link is updated #2263

StlMaris123:
In your PR did you:
Update case when link is updated #2263


rlgreen91:
The issue requires you to update the view and add a feature spec as well.

StlMaris123:
Hi @rlgreen91 , I am a bit lost on the content we are supposed to add on the view

rlgreen91:
@StlMaris123 so, on the issue I mention updating the case history page.  If you look at the history view, you'll see the text mentions that a new version isn't created when a link is added or updated.  You need to change that text.

Issue #2647, opened 2018-10-27
Add an endpoint to update email addresses


rlgreen91:
Add an endpoint in the user controller to update the user's email address.  This endpoint should be similar to the current `update` action, except that the only permitted parameter should be the email address.

PR #2648, opened 2018-10-28
Fix#2647 

rohan-techfreak:
Added endpoint for updating user email.
As mentioned in the #2647, this endpoint is similar to the current update action, expect that the only permitted attribute should be email.


rlgreen91:
@rohan-techfreak Thank you for your PR!  Can you change the endpoint name to be `update_email`?

rohan-techfreak:
@rlgreen91 Updated the endpoint to user_email as suggested.

PR #2655, opened 2018-10-28
Fix#1435 

rohan-techfreak:
rescue with 0.0 when there are no analytics


rlgreen91:
@rohan-techfreak Thanks for submitting a PR.  Unfortunately, this isn't really where we would like to go with this - if there isn't any data for analytics, then we want to do something a bit more proactive than just have a blank page.  Accordingly, I'll be closing this.  If you're still looking for issues to tackle as part of Hacktoberfest, please feel free to look through our issues labelled as such.

PR #2657, opened 2018-10-28
Remove two unused methods.

Anne97:
## Description
The two unused methods `mailbox` and `conversations` are to be removed in the application_controller.rb

## Issue addressed
The unused methods have been successfully removed.


Anne97:
@rlgreen91  I'm not sure why the CI fails.

Issue #2493, opened 2018-10-11
Update maps route


rlgreen91:
In routes.rb, update the maps route to be the following:

get '/maps', to: "maps#index"
You may need to update the corresponding tests.

manasbedmutha98:
I'd like to take it.

PR #2664, opened 2018-10-30
Update map routes #2493

prathamsharma92:
Change map index route


rlgreen91:
Thanks for your contribution @prathamsharma92

PR #2666, opened 2018-10-30
Changes in routes file for new methods

shashank76:
#2647


rlgreen91:
Thanks for your PR @shashank76!  However, someone else has already made this change via an earlier PR, so I will close this one.  If you're still looking for Hacktoberfest issues to work on, please check out the issues we have labelled as such.

PR #2703, opened 2018-11-22
2696 cases in redis cache

trystant:
In your PR did you:

  - [X] Include a description of the changes? Caches the maps query so that there is only one hit per redis server restart rather than a database request every time the page is loaded. 
  - [X] Mention the issue the PR addresses?  This will significantly reduce the page load times after the initial page load
  - [ ] Include screenshots of any changes to the UI? N/A


app/controllers/cases_controller.rb
```diff
@@ -114,7 +115,7 @@ def undo
       end
       flash[:success] = 'Undid that!'
       flash[:undo] = @this_case.versions
-    rescue
+    rescue StandardError
```
rlgreen91: Let's not rescue all of `StandardError`

trystant: OK. This can be modified to handle the specific sets of errors expected. Does that work better @rlgreen91 ?

rlgreen91: Yes, it does.

rlgreen91:
Is this good for me to merge @trystant ?

Issue #2459, opened 2018-09-25
Set up meta tag generation


rlgreen91:
Set up the structure of meta tag generation based on this article: https://www.lewagon.com/blog/setup-meta-tags-rails.

Follow all of the steps until adding code to the views.  In your .env file, set the host variable equal to localhost.

rlgreen91:
Either one of you can take it, I would just prefer for it to be done by Wednesday.

campgurus:
Checking in on the status of this. I want to start posting case links to Twitter soon, which will help for SEO.

PR #2737, opened 2018-12-22
for issue #2459 Social tags to add social meta tags to better format shared content

campgurus:
In your PR did you:


campgurus:
I removed the metamagic gem which has not been updated in 4 years in favor of the meta-tags gem. Let me know if I need to go back and make those changes in a separate PR.

rlgreen91:
I'm repeating what I said in Slack here - there's no need to use a gem for this logic.  This is something that we can write our own logic for that will be relatively easy to maintain.  Rachael's PR was about this and the issue, #2459 has instructions on how to do so.

PR #2743, opened 2019-01-03
Makes feature specs for testing following and unfollowing functionality more idiomatic.

rlgreen91:
In your PR did you:

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -24,20 +24,20 @@ class CasesController < ApplicationController
     @cases = Case.all.order('date DESC').includes(:state).page(params[:page]).per(page_size) if !params[:query].present? && !params[:state_id].present?
   end
 
+  # rubocop:disable Metrics/AbcSize
   def show
     @this_case = Case.includes(:comments, :subjects).friendly.find(params[:id])
     @comments = @this_case.comments
     @comment = Comment.new
     @subjects = @this_case.subjects
-    if user_signed_in?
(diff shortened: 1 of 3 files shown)


trystant:
I like the simplification. I wonder if the following logic:
```
 current_user.follows.find_by_followable_id(@this_case.id) 
``` 
could be made a `User`  instance method named `follows` that reads as follows:
```
def follows(this_case = nil)
  return nil if this_case.blank?
  self.following?(@this_case)
end 
```

 @follow_id = current_user.follows.find_by_followable_id(@this_case.id) if user_signed_in?
```

could be
```
 @follow_id = current_user.id if current_user.follows(@this_case) && user_signed_in?
```

Or, by using the `[follows_scoped](https://github.com/tcocca/acts_as_follower#acts_as_followable-methods)` method in `[acts_as_follower](https://github.com/tcocca/acts_as_follower)` perhaps.

Does this make sense?

rlgreen91:
That last statement isn't possible though - `follow_id` refers to the ID of the actual follow association, which is a `has_many, :through` type of association.   There's `follows_scoped`, but since it returns the scope you still need to add something to get the actual `follow` - something like `this_case.follows_scoped.find_by(follower_id: current_user.id)`.  To me, that, and in particular `follows_scoped` isn't as clear as what I put so I wouldn't be a fan of the change.

Honestly, I'd really like to see a `find_follow(follower:, followable:)` method that returns the follow association but that probably needs to happen at the gem level.

trystant:
@rlgreen91 I agree with these ultimately being best served in the gem; I didn't realize that `follow_id` was the id of the association.

Issue #2551, opened 2018-10-20
Add a policy for analytics


rlgreen91:
Add a policy `AnalyticsPolicy` that checks to see if the user is an admin or analyst when trying to access the index.  Then, in the `index` action for the analytics controller, use `authorize` to call the policy and rmoeve the private method that checks if the user is an admin or analyst - instead, just have a before action filter to authenticate the user.

For an example, see: https://github.com/varvet/pundit#policies

PR #2756, opened 2019-01-22
Add a policy for analytics #2551

StlMaris123:
In your PR did you:
Add a policy for analytics
- Solves  #2551


app/controllers/analytics_controller.rb
```diff
@@ -3,9 +3,9 @@
 # Controller for analytics page. A lot of reporting going on.
 # TODO: Some of these reports may need to be asynchronous as the data grows
 class AnalyticsController < ApplicationController
-  before_action :check_for_admin_or_analyst, only: %i[show index]
```
rlgreen91: Honestly, there is no show action so you can remove any references to that.

app/policies/analytic_policy.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# User Policy
+class AnalyticPolicy < Struct.new(:user, :analytic)
```
rlgreen91: Be sure to account for CodeClimate issues.  This includes using `#rubocop:disable` if needed.

app/controllers/analytics_controller.rb
StlMaris123: I actually remove that line completely because the `before_action` was doing exactly what the policy will be doing

rlgreen91:
Ah, I thought I could approve each change rather than all of them.  @StlMaris123 you still need to make the fix for CodeClimate complaining about the initializer.  If you need to disable the cop so that it ignores the check, that's fine.

app/policies/analytic_policy.rb
trystant: I'm a bit unclear about why the `Struct` is needed here.

StlMaris123: screenshot from 2019-01-28 15-34-57
I used because analytic does not have a model associated with it.

trystant: @StlMaris123 This makes sense. Thanks for the explanation

rlgreen91:
@StlMaris123 you'll need to rebase again.

PR #2800, opened 2019-02-19
Changes session store from cookie store to database

rlgreen91:
Please note that this will be deployed to prod during off-hours since the app will need to be restarted.


rlgreen91:
I don't really see the need to. I would keep in mind that we had already
installed a corresponding gem that most likely takes care of that. This
answer is older than other answers to this issue.

PR #2803, opened 2019-02-22
Adds partials for case thumbnails

rlgreen91:
Adds partials to render case thumbnails on agency and user show views and does minor refactoring. 

This also includes code to skip using Bullet for `users#show`.  If Bullet is enabled, then it will complain that we should use eager loading to retrieve the state for each case.  However, doing so is not currently possible given the configuration of users, cases, and states - we'd need to change that in a sense, and that's not the point of this PR.  Hence, a filter applied to only that controller.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -4,6 +4,7 @@
 class UsersController < ApplicationController
   def show
     @user = User.find(params[:id])
+    @cases = @user.all_following
   end
 
   def edit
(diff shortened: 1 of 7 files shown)


rlgreen91:
@trystant Basically.  This is what twice now where I've run into an issue because those methods return arrays, and not relations like you'd expect.  Considering that we use the most basic version of it, it's hard not to imagine writing it myself in EBWiki and make it easy to extend.

Issue #1353, opened 2018-03-14
Add an editors section to the analytics page


rlgreen91:
Add an editors section to the analytics page.  You can more or less follow the changes that have been made in PR #1201 - just note that a lot of the changes made to `analytics_helper` won't be necessary, since those queries are performed elsewhere.

PR #2837, opened 2019-02-27
Add editors to analytics

campgurus:
This PR adds a table in the analytics view to show the most recent 10 editors addressing issue #1353.

the new table looks like the screenshot below:

screen shot 2019-02-26 at 11 51 54 pm

I did not add any new specs. Maybe I could verify the helper methods?

In your PR did you:

### DIFF
--- a/app/controllers/analytics_controller.rb
+++ b/app/controllers/analytics_controller.rb
@@ -18,6 +18,8 @@ class AnalyticsController < ApplicationController
     @cases_sorted_by_followers = Case.sorted_by_followers 10
     @most_visited_cases = DetermineVisitsToCases.call(@visits_this_week.sorted_by_hits(13))
     @most_recent_comments = Comment.sorted_by_creation 15
+    @case_updates_30 = PaperTrail::Version.where('created_at > ?', 30.days.ago)
+    @case_updaters_count_10 = @case_updates_30.group(:whodunnit).order('count_id DESC').limit(10).count(:id)
   end
 
   private
--- a/app/helpers/analytics_helper.rb
+++ b/app/helpers/analytics_helper.rb
@@ -25,14 +25,4 @@ module AnalyticsHelper
   def link_to_comment_user(comment)
     link_to User.find(comment.user_id).name, user_path(User.find(comment.user_id))
   end
-
-  def case_updates
-    # returns all versions of all cases created in the last 30 days
-    PaperTrail::Version.where('created_at > ?', 30.days.ago)
-  end
-
-  def case_updaters_count(how_many)
-    # returns list of uniq editors of case updates along woth thenumber of updates each has authored
-    case_updates.group(:whodunnit).order('count_id DESC').limit(how_many).count(:id)
-  end
 end
--- a/app/helpers/analytics_helper.rb
+++ b/app/helpers/analytics_helper.rb
@@ -25,4 +25,14 @@ module AnalyticsHelper
   def link_to_comment_user(comment)
     link_to User.find(comment.user_id).name, user_path(User.find(comment.user_id))
   end
+
+  def case_updates
+    # returns all versions of all cases created in the last 30 days
+    PaperTrail::Version.where('created_at > ?', 30.days.ago)
+  end
+
+  def case_updaters_count(how_many)
+    # returns list of uniq editors of case updates along woth thenumber of updates each has authored
+    case_updates.group(:whodunnit).order('count_id DESC').limit(how_many).count(:id)
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -18,6 +18,10 @@ class User < ActiveRecord::Base
   extend FriendlyId
   friendly_id :slug_candidates, use: %i[slugged finders]
 
+  scope :recent_editors, -> {
+    where(recent_editor: true )
+  }
+
   def mailboxer_name
     name
   end
@@ -39,6 +43,18 @@ class User < ActiveRecord::Base
     elsif mailchimp_user.is_a?(Hash)
       mailchimp_user['status']
     end
+  end
+
+  def versions
+    PaperTrail::Version.where(whodunnit: id)
+  end
+
+  def recent_edits(period)
+    versions.where('created_at > ?', period.days.ago)
+  end
+
+  def recent_editor
+    versions.where('created_at > ?', 30.days.ago).count >=1
   end
 
   private
--- a/app/views/analytics/_comments.html.erb
+++ b/app/views/analytics/_comments.html.erb
@@ -1,0 +1,16 @@
+<div class="chart-wrapper">
+  <div class="chart-title background-gray">
+    Recent Comments
+  </div>
+  <ul class="scroller">
+    <% Comment.order(created_at: :desc).each do |comment| %>
+      <li>
+        <blockquote class="blockquote">
+          <p><%= link_to truncate(Case.find(comment.commentable_id).title, length: 30), case_path(Case.find(comment.commentable_id)) %> -- <%= comment.created_at.strftime("%B %d, %Y") %></p>
+          <p><%= link_to truncate(comment.content, length: 100), case_path(Case.find(comment.commentable_id)) %></p>
+          <small><%= link_to User.find(comment.user_id).name, user_path(User.find(comment.user_id)) %></small>
+        </blockquote>
+      </li>
+    <% end %>
+  </ul>
+</div>
--- a/app/views/analytics/_editors_list.html.erb
+++ b/app/views/analytics/_editors_list.html.erb
@@ -1,0 +1,30 @@
+<div class="chart-wrapper">
+  <div class="chart-title background-orange">
+    Most Recent Editors - Last 30 days <%= image_tag('help_icon.png', data: {:toggle => 'popover', :container => 'body', :trigger => 'hover',:placement => "bottom", :title => "Active Editors", :content => "This metric shows the number of cases added or updated by editors in the past 30 days."})%>
+  </div>
+  <div class="chart-stage">
+   <table class='table leader-list'>
+     <tr>
+      <th>Editor</th>
+      <th>Updates</th>
+     </tr>
+
+     <% @case_updaters_count_10.each do |who| %>
+       <% user_id = who[0].to_i %>
+       <% user_object = who[1] %>
+       <tr>
+         <% unless user_id == 0 %>
+           <td><%= @user = User.find(user_id).name %></td>
+           <td><%= link_to user_object, @user %></td>
+         <% else %>
+           <td> Anonymous </td>
+           <td><%= user_object %></td>
+         <% end %>
+       </tr>
+     <% end %>
+   </table>
+  </div>
+  <div class="chart-notes">
+    The ten most active editors over the last 30 days.
+  </div>
+</div>
--- a/app/views/analytics/_most_followed_cases.html.erb
+++ b/app/views/analytics/_most_followed_cases.html.erb
@@ -1,0 +1,23 @@
+<div class="chart-wrapper">
+  <div class="chart-title background-orange">
+    Most Followed Cases
+  </div>
+  <div class="chart-stage">
+   <table class='table leader-list'>
+     <tr>
+      <th>Case</th>
+      <th>Followers</th>
+     </tr>
+
+     <% @cases_sorted_by_followers.each do |this_case| %>
+       <tr>
+         <td><%= link_to_case_title(this_case, 25) %></td>
+         <td><%= link_to_case_followers(this_case) %></td>
+       </tr>
+     <% end %>
+   </table>
+  </div>
+  <div class="chart-notes">
+    These are the cases with the most followers
+  </div>
+</div>
--- a/app/views/analytics/_most_viewed_cases.html.erb
+++ b/app/views/analytics/_most_viewed_cases.html.erb
@@ -1,0 +1,25 @@
+<div class="chart-wrapper">
+  <div class="chart-title background-gray">
+    Most Viewed by New Visitors - Last 7 days <%= image_tag('help_icon.png', data: {:toggle => 'popover', :container => 'body', :trigger => 'hover',:placement => "bottom", :title => "Cases Viewed By New Visitors", :content => "This metric shows the number of new visitors landing on specific EBWiki cases."})%>
+  </div>
+  <div class="chart-stage">
+   <table class='table leader-list'>
+     <tr>
+      <th>Page</th>
+      <th>Views</th>
+     </tr>
+
+     <% @most_visited_cases.each do |case_info| %>
+      <% this_case = case_info[0] %>
+      <% views = case_info[1] %>
+      <tr>
+        <td><%= link_to_case_title(this_case, 25) %></td>
+        <td><%= views %></td>
+      </tr>
+     <% end %>
+   </table>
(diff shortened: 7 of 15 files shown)


campgurus:
The test that is failing here is passing on my local. Not sure what is not working with Travis

app/helpers/analytics_helper.rb
```diff
@@ -25,4 +25,14 @@ def link_to_comment(comment, length)
   def link_to_comment_user(comment)
     link_to User.find(comment.user_id).name, user_path(User.find(comment.user_id))
   end
+
+  def case_updates
+    # returns all versions of all cases created in the last 30 days
+    PaperTrail::Version.where('created_at > ?', 30.days.ago)
```
rlgreen91: Why are you doing database queries in helpers?  Those should be in the controller.

app/helpers/analytics_helper.rb
```diff
@@ -25,4 +25,14 @@ def link_to_comment(comment, length)
...
+  end
+
+  def case_updaters_count(how_many)
+    # returns list of uniq editors of case updates along woth thenumber of updates each has authored
+    case_updates.group(:whodunnit).order('count_id DESC').limit(how_many).count(:id)
```
rlgreen91: Same as above.

app/views/analytics/index.html.erb
```diff
@@ -147,26 +147,59 @@
...
+            Most Recent Editors - Last 30 days <%= image_tag('help_icon.png', data: {:toggle => 'popover', :container => 'body', :trigger => 'hover',:placement => "bottom", :title => "Active Editors", :content => "This metric shows the number of cases added or updated by editors in the past 30 days."})%>
+          </div>
+          <div class="chart-stage">
+           <table class='table leader-list'>
+             <tr>
+              <th>Editor</th>
+              <th>Updates</th>
+             </tr>
+
+             <% case_updaters_count(10).each do |who| %>
```
rlgreen91: Don't iterate through the array - that has a performance hit.  Find a way to render this as a collection via a partial.

app/views/analytics/index.html.erb
```diff
@@ -147,26 +147,59 @@
...
+               <% user_id = who[0].to_i %>
+               <% user_object = who[1] %>
+               <tr>
+                 <% unless user_id == 0 %>
```
rlgreen91: Why do we need this check here?

app/views/analytics/index.html.erb
```diff
@@ -147,26 +147,59 @@
...
+                   <td><%= @user = User.find(user_id).name %></td>
```
rlgreen91: Database queries need to go in the controller, not the views.  Plus, if you already have the user object, there shouldn't be a need to run a second query just to get their name - just pull it from the object.

app/views/analytics/index.html.erb
```diff
@@ -147,26 +147,59 @@
...
+                   <td><%= link_to user_object, @user %></td>
+                 <% else %>
+                   <td> Anonymous </td>
```
rlgreen91: How is this branch even possible if only signed in users can edit cases?

campgurus:
> Overall, in addition to these changes, I'd like to take a step back - what is the purpose of adding this table? Sure, we can see the top ten editors, but what will we actually do with that information?
> 
> In addition to what I mentioned here, you'll also need to get the tests passing on Travis.

Editing cases is a core behavior we are trying to encourage. Currently, we have no convenient way to see who is editing cases and thus no quick way to see how we are doing on that score. Likewise, at some point this information should be viewable by the public in the same way that Amazon encourages its most prolific reviewers by identifying and listing them publicly.

app/helpers/analytics_helper.rb
campgurus: I can move them.

app/helpers/analytics_helper.rb
campgurus: Will move.

app/views/analytics/index.html.erb
campgurus: Ok. I will try to find a tutorial and implement this.

app/views/analytics/index.html.erb
campgurus: IIRC, this check was to account for instances where the edit is made from the admin panel or the console. I will confirm.

app/views/analytics/index.html.erb
campgurus: Not sure how to do this. `case_updaters_count` doesn't return the user object, only the user ids corresponding to the whodunnits in order of the most prolific editors. 

To get around this, I think I can add a a model method that counts the number of edits by a user over the past XX days. Then in the controller, create the instance variable that calls the method and returns the top ten editors.

Does that make sense?

app/views/analytics/index.html.erb
campgurus: Cases are edited via the admin panel, rails console and rake tasks, where current_user is nil.

app/views/analytics/index.html.erb
trystant: @campgurus look through our codebase for `render partial`. There should be examples of what @rlgreen91 has discussed there.

app/helpers/analytics_helper.rb
trystant: Also, that query is going to blow up as the number of cases & versions gets larger. Is there a way to narrow the scope of this query?

app/views/analytics/index.html.erb
trystant: @campgurus I'd make it a model scope rather than a method, but that's a step in the right direction.

trystant:
Also, I feel that the sections that are distinguished by comments like:
`<!-- Comments List-->`
`<!-- Editors List-->`
`<!-- Most Followed cases -->`

should be in separate partials.

app/controllers/analytics_controller.rb
```diff
@@ -18,6 +18,8 @@ def index
     @cases_sorted_by_followers = Case.sorted_by_followers 10
     @most_visited_cases = DetermineVisitsToCases.call(@visits_this_week.sorted_by_hits(13))
     @most_recent_comments = Comment.sorted_by_creation 15
+    @case_updates_30 = PaperTrail::Version.where('created_at > ?', 30.days.ago)
```
rlgreen91: Hold on - if we're trying to display the last 10 editors, then why are we querying for the versions of cases created in the past 30 days?

app/controllers/analytics_controller.rb
```diff
@@ -18,6 +18,8 @@ def index
...
+    @case_updaters_count_10 = @case_updates_30.group(:whodunnit).order('count_id DESC').limit(10).count(:id)
```
rlgreen91: Let's answer the question above before we dissect this query.

app/models/user.rb
```diff
@@ -41,6 +45,18 @@ def mailchimp_status
     end
   end
 
+  def versions
+    PaperTrail::Version.where(whodunnit: id)
```
rlgreen91: This is a code smell - why does the user need to know to look for a version of a case, or how that process works? Keep in mind that at some point, we may want to add versioning to other entities like agencies, organizations, or events - the way it's written right now that won't work.

I like the idea of the user noting the last time they edited a case, but not like this.

app/views/analytics/_comments.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="chart-wrapper">
+  <div class="chart-title background-gray">
+    Recent Comments
+  </div>
+  <ul class="scroller">
+    <% Comment.order(created_at: :desc).each do |comment| %>
```
rlgreen91: Look at how I handled the case thumbnails on the homepage.  That's how you should implement this.

app/views/analytics/_editors_list.html.erb
```diff
@@ -0,0 +1,30 @@
+<div class="chart-wrapper">
+  <div class="chart-title background-orange">
+    Most Recent Editors - Last 30 days <%= image_tag('help_icon.png', data: {:toggle => 'popover', :container => 'body', :trigger => 'hover',:placement => "bottom", :title => "Active Editors", :content => "This metric shows the number of cases added or updated by editors in the past 30 days."})%>
```
rlgreen91: Same as above.

app/views/analytics/_editors_list.html.erb
```diff
@@ -0,0 +1,30 @@
...
+         <% else %>
+           <td> Anonymous </td>
+           <td><%= user_object %></td>
+         <% end %>
+       </tr>
+     <% end %>
+   </table>
+  </div>
+  <div class="chart-notes">
+    The ten most active editors over the last 30 days.
```
rlgreen91: Wait, this doesn't match the wording used above.  Is this the most recent editors, as in the last 10 people to edit a case?  Or is it the 10 people who made the most edits in the past 30 days?

app/views/analytics/_most_followed_cases.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="chart-wrapper">
+  <div class="chart-title background-orange">
+    Most Followed Cases
```
rlgreen91: See above about partials.

app/views/analytics/_most_viewed_cases.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="chart-wrapper">
+  <div class="chart-title background-gray">
+    Most Viewed by New Visitors - Last 7 days <%= image_tag('help_icon.png', data: {:toggle => 'popover', :container => 'body', :trigger => 'hover',:placement => "bottom", :title => "Cases Viewed By New Visitors", :content => "This metric shows the number of new visitors landing on specific EBWiki cases."})%>
```
rlgreen91: See above about partials.

app/views/analytics/_new_users_by_day.html.erb
```diff
@@ -0,0 +1,11 @@
+<div class="chart-wrapper">
+  <div class="chart-title background-gray">
+    New Users by Day - Last <%= @last_days %> days (<%= @users.count %>)
```
rlgreen91: This count query needs to go.

app/views/analytics/_new_users_by_day.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  </div>
+  <div class="chart-stage">
+    <%= column_chart metric_grouped_by_day(@users, :created_at).count %>
```
rlgreen91: Same as above.

app/views/analytics/_recently_updated_cases.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    Recently Updated Cases
+  </div>
+  <div class="chart-stage">
+   <table class='table leader-list'>
+     <tr>
+      <th>Case</th>
+      <th>Updated</th>
+     </tr>
+
+     <% @cases_sorted_by_update.each do |this_case| %>
```
rlgreen91: See comment above for rendering partials.

app/controllers/analytics_controller.rb
campgurus: The query is intended to find versions (new cases or edits of old cases) created in the past 30 days. Then group by whodunnit and lastly find those users and sort them in order of how many versions each of them is responsible for.

app/controllers/analytics_controller.rb
campgurus: so the case_updaters_count_10 query takes all the versions from the past 30 days and groups them by the whodunnit (id of the responsible user), returning the id and the number of versions it "has"

app/views/analytics/_comments.html.erb
campgurus: Ok thanks. I hadn't gotten to these partials yet.

app/models/user.rb
campgurus: This method doesn't actually get used at this point. I was thinking of how to move the new controller instance variables into model scopes and thought this would be helpful for that purpose, but decided against it. Then I was thinking it would eventually be useful to show on the user profile page what cases a person has edited. I will just remove it though.

app/controllers/analytics_controller.rb
trystant: This group of queries needs to be in another object that the `AnalyticsController` calls, I'm thinking.

rlgreen91:
@campgurus are you going to change those other sections into partials?  If not, I can do that in a separate PR.

campgurus:
> @campgurus are you going to change those other sections into partials? If not, I can do that in a separate PR.

yes. i will try to do it tomorrow afternoon.

app/controllers/analytics_controller.rb
rlgreen91: Honestly,  I think it might be better to have a column on the user that records the last time they created or updated a case.  Then, it's just a matter of querying for that column rather than pulling what feels like an arbitrary number of versions and running a group query on it.

rlgreen91:
> > @campgurus are you going to change those other sections into partials? If not, I can do that in a separate PR.
> 
> yes. i will try to do it tomorrow afternoon.

Ok.  It might be easier to just do that in a separate PR so it doesn't get lost or held up in the midst of this.

campgurus:
> > > @campgurus are you going to change those other sections into partials? If not, I can do that in a separate PR.
> > 
> > 
> > yes. i will try to do it tomorrow afternoon.
> 
> Ok. It might be easier to just do that in a separate PR so it doesn't get lost or held up in the midst of this.

Ok. May be better like you said a while back to keep PRs smaller. If you want to break it down in to smaller pieces, that would definitely be easier.

rlgreen91:
@campgurus took a look at the example for `calculate` in this doc: https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html.  Particularly, look at how they use `group` in combination with `maximum`.  We could probably use `group` and maybe `having` in combination with other methods to simplify the query.

rlgreen91:
> > > > @campgurus are you going to change those other sections into partials? If not, I can do that in a separate PR.
> > > 
> > > 
> > > yes. i will try to do it tomorrow afternoon.
> > 
> > 
> > Ok. It might be easier to just do that in a separate PR so it doesn't get lost or held up in the midst of this.
> 
> Ok. May be better like you said a while back to keep PRs smaller. If you want to break it down in to smaller pieces, that would definitely be easier.

Wait, to be clear, do you want me to do this, or you?

campgurus:
> > > > > @campgurus are you going to change those other sections into partials? If not, I can do that in a separate PR.
> > > > 
> > > > 
> > > > yes. i will try to do it tomorrow afternoon.
> > > 
> > > 
> > > Ok. It might be easier to just do that in a separate PR so it doesn't get lost or held up in the midst of this.
> > 
> > 
> > Ok. May be better like you said a while back to keep PRs smaller. If you want to break it down in to smaller pieces, that would definitely be easier.
> 
> Wait, to be clear, do you want me to do this, or you?

I can break it down, but I would probably do better to get your guidance on what the appropriate pieces are. If you list them here, I could make issues and then go through them one by one and make prs for each.

rlgreen91:
@trystant @campgurus  Honestly, enough changes have been made to both the app as a whole and the analytics page that I recommend closing this PR and starting on a new, update branch.  Then those changes can more closely align with the way we've structured those other charts.

trystant:
@rlgreen91 Shutting this down & taking it from the top?

Issue #1301, opened 2018-02-28
Refactor analytics code from Article/Case


trystant:
The following methods in the `Article` or `Case` model are not related to the model directly, but analytics:

```
mom_new_cases_growth
mom_cases_growth
cases_updated_last_30_days
mom_growth_in_case_updates
```

This should be in a separate concern

rlgreen91:
Agreed.  I would like to extract similar methods located elsewhere as well.

PR #2867, opened 2019-03-10
Refactors case statistical methods

rlgreen91:
Directly addresses Issue #1301. Refactors methods for calculating percentage growth statistics for cases.  Moves statistical methods for cases to separate statistics module.  Refactors analytics controller and dashboard using statistics module.

To be clear, there is a lot to do with regard to refactoring these methods that appear in multiple models.  On top of that, there are enhancements we could do like using value objects or performing some of these calculations at the database level.  But, this is a good first step.

In your PR did you:


spec/lib/statistics_spec.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Statistics do
+
+  def mom_changed_cases_growth(start_cases:, end_cases:)
```
trystant: Seems like these could be helpful in the `Statistics` model itself, not just in the specs

spec/lib/statistics_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+require 'rails_helper'
+
+RSpec.describe Statistics do
+
+  def mom_changed_cases_growth(start_cases:, end_cases:)
+    Statistics.mom_changed_cases_growth(cases_at_start: start_cases, cases_at_end: end_cases)
+  end
+
+  def mom_cases_growth(start_cases:, end_cases:)
+    Statistics.mom_cases_growth(cases_at_start: start_cases, cases_at_end: end_cases)
```
trystant: What happens if an invalid date or other data types are sent as arguments?

spec/lib/statistics_spec.rb
rlgreen91: What, to basically have instance methods that act as a shell for class methods.  I'm not so sure if that's a good idea - I like being explicit about what module the method is coming from.

spec/lib/statistics_spec.rb
rlgreen91: Yeah, I have that as one of my things to tackle as part of this.  It's technically always been a concern, but everything was so hard coded that we didn't have to care.

rlgreen91:
@trystant I can't get rid of that carriage return it's crying about but otherwise it's good.

trystant:
@rlgreen91  You tried `rubocop -a`?

rlgreen91:
I'm worried it's going to change all of the files, and I'll have to clean that up from a git standpoint.

trystant:
You can mention `rubocop -a <filename>` with a filename so it doesn't do the whole repo

Issue #2851, opened 2019-03-02
Reduce cyclomatic complexity for mom_visits_growth


trystant:
This method is too complex:
```
def mom_visits_growth
    last_month_visits = Ahoy::Visit.this_month.count
    return 0 if last_month_visits.zero?

    last_60_days_visits = Ahoy::Visit.most_recent(60.days.ago).count
    prior_30_days_visits = last_60_days_visits - last_month_visits
    return (last_month_visits * 100) if prior_30_days_visits.zero?

    (((last_month_visits.to_f / prior_30_days_visits) - 1) * 100).round(2)
  end
```

There needs to be a way to simplify this functionality as well as make it testable

trystant:
It's not just me; CodeClimate says so
![screenshot 2019-03-02 at 06 59 58](https://user-images.githubusercontent.com/13509/53681519-dbc0b780-3cb8-11e9-875b-82c81fc94595.png)

rlgreen91:
CodeClimate is a rotten liar!!!!

PR #2883, opened 2019-03-23
Add follow and visit methods to statistics module

rlgreen91:
Moves statistical methods from follow and visit models to statistics modules. Adds query object to determine count of distinct followers. Refactors statistical methods.  Addresses #2851 

In your PR did you:


rlgreen91:
Now with query objects!  This will allow us to have a place to stick complicated queries.  While `distinct_count` could be a scope, I prefer for those to be extensible, and therefore should return a relation.  In theory query objects should return relations to but I'm not as opposed to the idea of them not returning one.

I think after this I'll play around with value objects for dates and date ranges to reduce the amount of hardcoding. 

Technically we should allow negative percentages when calculating growth - was there ever a reason why we didn't?

trystant:
My guess is that we didn't plan for negative percentages because the stats _should always be positive!_

Looking forward to checking out the query objects.

rlgreen91:
> My guess is that we didn't plan for negative percentages because the stats _should always be positive!_
> 
> Looking forward to checking out the query objects.

Ok, but that doesn't necessarily reflect reality.  Do you want me to update the code to allow negative percentages @trystant?

lib/statistics.rb
```diff
@@ -2,23 +2,35 @@
...
 # This module contains a number of methods to calculate various statistics
 module Statistics
-  def self.mom_changed_cases_growth(cases_at_start:, cases_at_end:)
-    return 0 if cases_at_end.size.zero?
+  def self.mom_changed_metric_growth(metric_at_start:, metric_at_end:)
+    return 0 if metric_at_end.size.zero?
 
-    case_growth = cases_at_start.size - cases_at_end.size
-    return (cases_at_end.size * 100) if case_growth.zero?
+    metric_growth = metric_at_start.size - metric_at_end.size
```
trystant: It looks like lines 8 - 9, 17 & 18 and 30, 31 can be refactored into another method called 
`def self.metric_growth(metric_start, metric_end)
  growth = metric_start.size - metric_end.size
  (growth.zero?) ? metric_end.size * 100 : growth 
end
```

I think it can help abstract that out to a separate function to see how the others exist with the refactored form.

trystant:
> Ok, but that doesn't necessarily reflect reality. Do you want me to update the code to allow negative percentages @trystant?

Yes please.

rlgreen91:
@trystant everything is all fixed now.  It still complains about one method being too long, despite me disabling the cop for it.  That error disappears - on a couple of builds CodeClimate has marked it as fixed.

trystant:
One thing I might do is disable that code climate rule but out isn't a blocker

Issue #1798, opened 2018-07-18
Add mailer previews (3 of 4)


rlgreen91:
Note: This issue is 3 of 4 for the given topic, and must be followed in the order listed. The preceding issues must be completed and merged before continuing to the next issue.

Create a new directory `spec/mailer/previews` and add previews for each of the mailers.  The previews should follow the format outlined here: http://guides.rubyonrails.org/action_mailer_basics.html.

rlgreen91:
We're going to make some other changes to how the mailers are organized, so I'll close this in favor of creating those issues.

PR #2909, opened 2019-03-30
Add mailer previews

rlgreen91:
Yeah, so this PR ended up being a lot bigger than I anticipated.  I originally went to work on issue #1798, by adding email previews.  However, in doing that, I found myself asking a lot of questions about why the code was organized the way it was, what we were trying to communicate, and ultimately a discovery that we weren't even sending this email to begin with.  So, this became basically an overhaul to add this functionality for sending emails when an admin is added in - which is why this PR touches almost 20 files.  Some of those are deleted files, and just configuration settings in different environments, but the bulk of it is code, and just as importantly, specs to verify that code works.

Next step after this will be to add an email when a user has their admin status removed.

----------------------------------------------------------------------------------------------------------------
In your PR did you:


rlgreen91:
@trystant I'm not entirely sure why that test fails on Travis but not locally - I think it may be a config issue.  I've asked Travis to enable debug mode on the repository so I can ssh into the build and see what's going on.

Issue #2858, opened 2019-03-07
Remove truncate method in user show view


rlgreen91:
In `app/views/users/show.html.erb`, don't use `truncate` when displaying the user's name - just display the name in full.

PR #2914, opened 2019-04-01
Remove truncate method in user show view

StlMaris123:
In your PR did you:
Remove truncate method in user show view
- Solves #2858


app/views/users/show.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="row">
       <div class="col-md-2">
         <%= image_tag avatar_url(@user,200),  class: "img-rounded img-responsive" %>
-        <%= truncate(@user.name, length:16) %>
+        <%= @user.name, length:16 %>
```
rlgreen91: You don't need the length parameter, just put the user's name.

PR #2922, opened 2019-04-06
Adds jurisdiction as postgres enum type.

rlgreen91:


trystant:
So the `db structure.sql` is a generated file. I don't think we need to include that

rlgreen91:
@trystant `structure.sql` is generated in place of `schema.rb`, so it should be included.

PR #2935, opened 2019-04-07
Adds ability for admins to make other users admins.

rlgreen91:
Basically, this PR is focused on making it easy for admins to change the admin status of another user.  We currently have a mailer that should be sent whenever a user becomes an admin.  But because of how Rails Admin is designed, that email never actually gets sent.  This allows an admin to change a user's admin status via the standard actions available through a user profile.  As part of that this includes:
- Conditionally showing the "edit profile" button if it is the user's profile or the user is an admin
- Restricting who can see the form generated under the `edit` action: because this form exposes the user's email address, it should really be limited in access to the user and admins
- Shows different forms: if user is editing their own profile, it shows the normal form, but if the user is an admin, it shows a different form with the admin checkbox and the rest of the fields as read-only

This PR also adds labels to the forms - they should be there for accessibility.  It also removes the request specs for the user endpoints - something odd is happening with database transactions.  It's not clear what is occurring, and it causes even existing specs to fail - it's not the point of this PR so I removed the file in question.


rlgreen91:
@trystant I am not.  I added this note to the warning: Authentication and authorization restrictions updates of admin status to other admins.  Specifically, the only way in which a user can access the admin parameter is if they are an admin.

trystant:
Looks like the test coverage for the `users_controller` went down a fair amount. Let's add tests to improve that coverage?

rlgreen91:
@trystant That's because I removed the request specs I added in the previous PR - for some reason it kept hanging on an invalid DB transaction, regardless of the test within the file.  No other spec had this issue.

PR #2950, opened 2019-04-13
Adds value object for date ranges.  Cleans up analytics controller.

rlgreen91:
In your PR did you:


rlgreen91:
One last thing I'd like to do - I'm not a fan of the term mom, because it's really not clear what that means.  I'd prefer to just say something like "50% increase*" with a note at the bottom that we measure that over the past 30 days.

Issue #2956, opened 2019-04-17
Convert categories to cause of death


rlgreen91:
In the codebase, we use categories to represent cause of death - this isn't terribly clear, so I'll change it to reflect the latter.  This issue will include:

- [ ] Rename category table to cause of death
- [ ] Rename model, spec, and all associated methods
- [ ] Rename controller, routes, and all associated methods
- [ ] Rename views, stylesheets, helpers, and all associated methods

Since this is mostly used in the context of a case, it won't be as extensive as renaming that was.

rlgreen91:
The case form always used the term "cause of death"

PR #2966, opened 2019-04-25
Redo categories

rlgreen91:
In your PR did you:

This PR addresses #2956, which is converting the category model to cause of death.  This also includes a minor change to the state factory.

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -46,7 +46,10 @@ class CasesController < ApplicationController
       flash[:undo] = @this_case.versions
       redirect_to @this_case
     else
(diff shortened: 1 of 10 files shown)


rlgreen91:
> I feel like `set_instance_vars` could become a pattern for all controllers if it hasn't already been done in an external gem. I like the semantic change of 'categories' to 'causes_of_death'.

I'm actually not really a fan of setting instance variables like that in controllers because it can make it difficult to keep track of which action sets it and how.  That's actually what happened when the history table "disappeared" - the wrong value was set in private methods and so `nil` was always passed as the history for a case.  But this has so many instance variables for a form that I'm willing to make an exception.

PR #3007, opened 2019-05-18
Conditionally show edit profile button

rlgreen91:
Adds javascript functionality to conditionally show edit profile button. Removes edit button partial.

In your PR did you:


rlgreen91:
no_edit_button
with_edit_button

app/assets/stylesheets/users.scss
```diff
@@ -78,3 +78,7 @@ label{
 .hidden-label{
 	display:none;
 }
+
+#edit-button {
+	display: none;
```
trystant: I think it may help to set the `visibility` property for elements we want to hide. There's more detail in this [stackoverflow](https://stackoverflow.com/questions/20598568/how-to-display-and-hide-a-div-with-css#24055535)

rlgreen91:
> I think it may help to set the `visibility` property for elements we want to hide. There's more detail in this [stackoverflow](https://stackoverflow.com/questions/20598568/how-to-display-and-hide-a-div-with-css#24055535) post.

@trystant I'm not sure I agree with that.  The key part is that `visibility: hidden` will preserve the space while `display: none` will not.  But why would we want to preserve the space in this case?

rlgreen91:
@trystant  I actually had to update the logic in this PR but that has been done and it's based off of recent master, so it's good to be reviewed.

PR #3051, opened 2019-06-22
Update to Rails 5

rlgreen91:
First update to Rails 5


app/assets/stylesheets/cases.scss
```diff
@@ -263,13 +263,23 @@ $color-dark-gray: #4D4D4D;
 }
 
 #unfollow-button {
-  // font-size: small;
   color: white;
-  background-color: $color-gray;
+  background-color: $color-orange;
```
trystant: @rlgreen91 Is this a consolidation of CSS files? I feel like it may be helpful to separate this into another PR in order to keep things manageable.

app/views/cases/_button_group.html.erb
```diff
@@ -1,13 +1,13 @@
 <div class = "btn-group btn-group-justified">
 	<% if user_signed_in? %>
 		<% if current_user.following?(@this_case) %>
-		    <%= link_to 'Unfollow', case_follow_path(@this_case, @follow_id), :confirm => "Are you sure you want to unfollow this case?", :method => :delete, class: "btn btn-primary.outline", id: "unfollow-button" %>
+		  <%= button_to 'Unfollow', follows_case_path(@this_case), method: :delete, data: { confirm: "Are you sure you want to unfollow this case?" }, form_class: "btn btn-primary btn-lg", id: "unfollow-button" %>
```
trystant: @rlgreen91 `button_to` has been around since 4.2. Is this an essential piece of upgrading to 5.0, or is this a separate kind of upgrade?

app/assets/stylesheets/cases.scss
rlgreen91: This is part of the situation I outlined in the comment below.

app/views/cases/_button_group.html.erb
rlgreen91: Part of upgrading.  Some of the changes I made ended up affecting how we handled the path for following cases.  That, in turn, led me to making the change to the path here, where I realized that eh, technically we shouldn't be using a link styled as a button to perform a post - it should actually be a button.  So I changed that here as part of it since I was already there.

Ironically enough, it was trying to fix the styling of the "unfollow" button that prompted my idea about the CSS fix.  The other two were fairly straightforward to change.

rlgreen91:
I'm not sure why I can't respond directly to the spec comment, but I'll do so here.  First, I think 21 request specs for 38 controller actions is about 55% yes, but keep in mind that we do have some non-rest controller actions so that number is probably lower.

Ideally, I'd like to get to 90+% coverage of controller actions, which would account for static pages that don't really need to be tested.  However, as we saw with the conversations request spec, adding them for some controller actions leads to issues and bug fixes that are more complicated than expected and most likely out of scope of this PR.

I would like to keep adding request specs, of course, but honestly, I think what we have here is fine as far as upgrading to 5 is concerned.  Given that our previous controller tests didn't cover everything and were often testing only a single happy path, I don't think we've lost much coverage in terms of functionality.

trystant:
@rlgreen91 I'd do one `rubocop -a` on the branch, update from master and then this PR will be good to go.

rlgreen91:
As per @trystant's message in Slack, this PR has been approved.

PR #3132, opened 2019-08-10
Fix case date validation

rlgreen91:
This PR reverts previous changes to make the datepicker default to the MM-DD-YYYY format.  For whatever reason, I am unable to change the format of the input data, so even if the date is inputted in the MM-DD-YYYY format, the app reads the data as DD-MM-YYYY.  Future changes to this form will change this field entirely.


rlgreen91:
> I tried to reverse this & ran into the same problem. I think that making that format change should be in another ticket.

Yeah, I'm working on doing that form using the built-in form builder in general, and part of that will be accounting for a MM-DD-YYYY format for the date.

Issue #3005, opened 2019-05-18
Remove `cause_of_death` model and table


rlgreen91:
Remove the `cause_of_death` model and table.  Note that this will include removing the association within the `Case` model and the corresponding conversion rake task in `cases.rake`.

amitech:
@rlgreen91 I can work on it. Please assign it to me.

PR #3153, opened 2019-08-25
Adds cause of death enum

rlgreen91:
Adds migration to create cause of death enum.  Adds rake task to convert cause of death reference data on case table to enum.

Addresses Issue #3005 in part

In your PR did you:

### DIFF
--- a/app/models/case.rb
+++ b/app/models/case.rb
@@ -5,6 +5,17 @@
 class Case < ApplicationRecord
   # TODO: Clean up relationship section
 
+  enum cause_of_death_name: {
+    choking: 'choking',
+    shooting: 'shooting',
+    beating: 'beating',
+    taser: 'taser',
+    vehicular: 'vehicular',
+    medical_neglect: 'medical neglect',
+    response_to_medical_emergency: 'response to medical emergency',
+    suicide: 'suicide'
+  }
+
   belongs_to :user
   belongs_to :cause_of_death
   belongs_to :state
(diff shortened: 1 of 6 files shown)


rlgreen91:
@trystant I can't for the life of me figure out why Rubocop is ignoring that I disabled the class length cop.  Either way, once we finish converting the model to an enum, that flag is likely to go away as other lines are deleted.

app/models/case.rb
```diff
@@ -2,9 +2,21 @@
 
 # This is the main model of the application. Each death is a case.
 # TODO: Lots & lots of refactoring
+# rubocop:disable Metrics/ClassLength
 class Case < ApplicationRecord
   # TODO: Clean up relationship section
```
trystant: This todo is from 2016. I think that that comment can be removed.

db/migrate/20190824175709_add_cause_of_death_to_postgres.rb
```diff
@@ -0,0 +1,17 @@
+class AddCauseOfDeathToPostgres < ActiveRecord::Migration[5.0]
+  def up
+    execute <<-SQL
+      CREATE TYPE cause_of_death AS ENUM ('choking', 'shooting', 'beating', 'taser', 'vehicular', 'medical neglect', 'response to medical emergency', 'suicide')
+    SQL
+
+    add_column :cases, :cause_of_death, :cause_of_death, index: true
+  end
+
+  def down
```
trystant: I like that this is completely reversible.

app/models/case.rb
rlgreen91: Ok, I'll remove both of them in the next PR touching this file.

PR #3165, opened 2019-09-02
Changes case controller and form to use cause of death enum

rlgreen91:
Changes case controller and form to use cause of death enum instead of model.

Addresses Issue #3005 in part.  Screenshot attached.

In your PR did you:

### DIFF
--- a/app/views/cases/_form.html.erb
+++ b/app/views/cases/_form.html.erb
@@ -37,7 +37,7 @@
 				<h4>Add a YouTube or Vimeo video url relating to the case:</h4>
 				<%= f.input :video_url %>
 				<h4>Cause of death:</h4>
-				<%= f.collection_select :cause_of_death_id, @causes_of_death, :id, :name, {prompt: "Choose a cause of death."} %>
+				<%= f.input :cause_of_death_name, collection: [:choking, :shooting, :beating, :taser, :vehicular, :medical_neglect, :response_to_medical_emergency, :suicide], label: false %>
 	    </div>
 		</fieldset>
 	</div>
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -128,7 +128,7 @@ class CasesController < ApplicationController
                                   :litigation,
                                   :community_action,
                                   :agency_id,
-                                  :cause_of_death_id,
+                                  :cause_of_death_name,
                                   :date,
                                   :state_id,
                                   :city,
--- a/config/locales/simple_form.en.yml
+++ b/config/locales/simple_form.en.yml
@@ -19,6 +19,16 @@ en:
           federal: 'Federal'
           university: 'University'
           commercial: 'Commercial'
+      case:
+        cause_of_death_name:
+          choking: 'Choking'
+          shooting: 'Shooting'
+          beating: 'Beating'
+          taser: 'Taser'
+          vehicular: 'Vehicular'
+          medical_neglect: 'Medical Neglect'
+          response_to_medical_emergency: 'Response to Medical Emergency'
+          suicide: 'Suicide'
     labels:
       agency:
         state: 'State'


rlgreen91:
Screenshot (19)

app/views/cases/_form.html.erb
```diff
@@ -37,7 +37,7 @@
 				<h4>Add a YouTube or Vimeo video url relating to the case:</h4>
 				<%= f.input :video_url %>
 				<h4>Cause of death:</h4>
-				<%= f.collection_select :cause_of_death_id, @causes_of_death, :id, :name, {prompt: "Choose a cause of death."} %>
+				<%= f.input :cause_of_death_name, collection: [:choking, :shooting, :beating, :taser, :vehicular, :medical_neglect, :response_to_medical_emergency, :suicide], label: false %>
```
trystant: Is there any advantage to listing these out explicitly as opposed to `CauseOfDeath.all`?

rlgreen91: > Is there any advantage to listing these out explicitly as opposed to `CauseOfDeath.values.all`?

The values are lowercase and have underscores instead of spaces.  Plus the model doesn't exist anymore, you'd have to use `Case`, unlike this which just accesses the locale file.

trystant: Seems like this would be a problem for line 161 in `app/controllers/cases_controller.rb`:
`
  @causes_of_death = SortCollectionOrdinally.call(collection: CauseOfDeath.all)
`

rlgreen91: No, it's not.

rlgreen91: Let me clarify - the model itself still exists.  I can't get rid of that until we update the data - which I'll do after this is pushed to prod.  So this line will still exist, and in the next PR for this whole issue I'll remove it.

trystant: Maybe there's something I'm missing.  I changed over to the `update_cause_of_death_field` branch and migrated, and:
```
➜  ebwiki git:(update_cause_of_death_field) ✗ rake db:migrate
== 20190824175709 AddCauseOfDeathToPostgres: migrating ========================
-- execute("      CREATE TYPE cause_of_death AS ENUM ('choking', 'shooting', 'beating', 'taser', 'vehicular', 'medical neglect', 'response to medical emergency', 'suicide')\n")
   -> 0.0244s
-- add_column(:cases, :cause_of_death, :cause_of_death, {:index=>true})
   -> 0.0073s
== 20190824175709 AddCauseOfDeathToPostgres: migrated (0.0321s) ===============

== 20190825000337 RenameCauseOfDeathForCases: migrating =======================
-- rename_column(:cases, :cause_of_death, :cause_of_death_name)
   -> 0.0164s
== 20190825000337 RenameCauseOfDeathForCases: migrated (0.0167s) ==============

➜  ebwiki git:(update_cause_of_death_field) ✗ bundle exec rails console
Loading development environment (Rails 5.0.7.2)
irb: warn: can't alias context from irb_context.
2.5.5 :001 > CauseOfDeath.all
  CauseOfDeath Load (0.9ms)  SELECT "causes_of_death".* FROM "causes_of_death"
 => #<ActiveRecord::Relation []>

Comment (rlgreen91) [reply app/views/cases/_form.html.erb]: Did you drop and reseed the database?  I don't think we have any records for cause of death there.  I can tell you that, when testing out my rake task, I verified that I do have records for the `CauseOfDeath` model.

Comment (trystant) [reply app/views/cases/_form.html.erb]: Records are also in staging and production. I can accept that as a "my machine" problem. I did some more tinkering and came up with this:
```
2.5.5 :013 > Case.causes_of_death_name.keys.map(&:to_sym).sort
 => [:beating, :choking, :medical_neglect, :response_to_medical_emergency, :shooting, :suicide, :taser, :vehicular] 
```
Mapping the Rails enum to a Postgres enum is truly a more elegant solution than the model solution that currently exists. There's something about manually listing those causes that seem to be potentially problematic. Whether it's done directly in this manner or a helper method that delegates to the logic I mentioned, seems like it would be better to maintain that elegance and link directly to the enum.

PR #3184, opened 2019-09-16
Removed gotcha from user registration process #3168

trystant:
In your PR did you:
Removed gotcha from the user registration process because it doesn't work with Rails 5 and Devise. Provides a short term fix for 3168

### DIFF
--- a/spec/requests/registrations_spec.rb
+++ b/spec/requests/registrations_spec.rb
@@ -1,0 +1,18 @@
+RSpec.describe 'Registrations', type: :request do
+  describe 'POST /registrations' do
+    it 'works! (now write some real specs)' do
+      get '/'
+      expect(response).to render_template('index')
+      
+      post user_registration_path, user: {
+          name: 'Mark Nyon',
+          email: 'here@there.com',
+          password: '12344321',
+          password_confirmation: '123443211',
+          subscribed: 0
+      }
+      expect(response).to have_http_status(200)
+    end
+  end
+end
+
--- a/spec/requests/registrations_spec.rb
+++ b/spec/requests/registrations_spec.rb
@@ -1,9 +1,11 @@
+# frozen_string_literal: true
+
 RSpec.describe 'Registrations', type: :request do
   describe 'POST /registrations' do
     context 'on success' do
       before do
         post user_registration_path, params: {
-            user: {
+          user: {
             name: 'Mark Nyon',
             email: 'here@there.com',
             password: '12344321',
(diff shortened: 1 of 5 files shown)


spec/requests/registrations_spec.rb
```diff
@@ -0,0 +1,18 @@
+RSpec.describe 'Registrations', type: :request do
```
rlgreen91: I would prefer this match the formatting of the other request specs.

spec/requests/registrations_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  describe 'POST /registrations' do
+    it 'works! (now write some real specs)' do
+      get '/'
+      expect(response).to render_template('index')
```
rlgreen91: I don't see a need for this assertion.

trystant: It was to confirm a request beforehand...but I can keep things to single spec per request.

spec/requests/registrations_spec.rb
trystant: Can you be specific? I modeled this after `spec/requests/organizations_spec.rb`

rlgreen91: Ah, it wasn't clear to me what the test was focused on before, but this looks good.

trystant:
> I still want to know why we're adding this gem back.

That gem is required for the `render_template` method used in the spec.

rlgreen91:
> > I still want to know why we're adding this gem back.
> 
> That gem is required for the `render_template` method used in the spec.

I see.  I'm not a fan of checking that the template is rendered in my request specs - I'd just test that some string in the body is present like you're doing now and leave it at that.  In my opinion, a request spec shouldn't really be concerned about what the response looks like in terms of HTML, just that the request was processed properly, and in this case that's including an error message in the body.  I don't particularly need to know that it rendered a specific template - that's what the view specs are for.

trystant:
I've read through documentation discussing the recommendation to remove `rails-controller-testing` and I don't think I agree with it, but my objection isn't strong enough to stop things at this point in time.

Why? I feel like it's helpful to know whether the request with params goes to the correct page. `render_template` was the mechanism that request specs used to determine that the correct page was rendered at the end of the request. Seems like it's possible to render a page that has the correct status, but incorrect location.

Using `pry` I noticed that both `response.uri` and `response.location` are `nil` in these tests, so that would probably have to be changed in order to make the expectation I'm looking for possible.

We'll see how it works though. I've updated specs and removed that gem from the build for now.

PR #3208, opened 2019-09-29
Changed find method & removed conditional statement

### LINKED ISSUE #3206: Change find method in case history action
In the history action in the cases controller, change the find method from `find_by_slug` to `find`.  This should also allow you to remove the check for `this_case.blank?` in the conditional.

birtony:
Fix #3206: changed find method in case history action from `find_by_slug` to `find` and removed `this_case.blank?` from conditional statements.


rlgreen91:
Thanks for your submission @birtony!

Issue #3197, opened 2019-09-27
Remove try method around versions in case history


rlgreen91:
In the history action in `cases_controller`, remove the try method call around versions - specifically, you should be able to call `@this_case.versions.order(created_at: :desc)`.  You can also remove the check for `@this_case.versions.blank?`

gkpacker:
Hey @rlgreen91, can i work on this?

PR #3245, opened 2019-10-02
[Fixes] Remove try method around versions in case history #3197

mfcruz:
remove try clause in history function replaced with rescue

In your PR did you:

  - [* ] Include a description of the changes?
  - [*] Mention the issue the PR addresses?


mfcruz:
> Thanks for your contribution @mfcruz! There are two issues to tackle - the first is that there's an extra `rescue` at the end of line 90. Additionally, the `rescue` block on line 91 should be aligned with `def` and `end`.

the rescue in line 90 will ensure that `@case_history` will have a blank array in case no versions are found

rlgreen91:
Thanks for your contribution @mfcruz! 

> > Thanks for your contribution @mfcruz! There are two issues to tackle - the first is that there's an extra `rescue` at the end of line 90. Additionally, the `rescue` block on line 91 should be aligned with `def` and `end`.
> 
> the rescue in line 90 will ensure that `@case_history` will have a blank array in case no versions are found

Right, so in our project, `#versions` comes from the paper-trail gem, which records a version for each creation and every change to a record (in our project, a case).  The only way there is no associated versions is if there is no case - so it should get caught via the normal 404.  If we do somehow have a case with no version, I'd rather it bubble up the error rather than catching it and returning an empty array.  That's why I'd prefer to drop that `rescue []`

rlgreen91:
Hi @mfcruz, are you going to be able to finish this soon?

rlgreen91:
This issue has been addressed in another PR, and so this one will be closed.

Issue #966, opened 2017-07-30
Redirect user after attempt to sign up with a previously used email


rlgreen91:
If a user tries to sign up for the site using an email that is already used for another account, redirect them to the sign in page.

luhova:
Hello, @rlgreen91 . Last year I contributed to your project as part of the Hacktoberfest and now I would like to do it again :) You can assign me this issue if it's not done yet.

PR #3291, opened 2019-10-08
redirect on duplicate email

Dfenniak:
This PR is meant to address https://github.com/EBWiki/EBWiki/issues/966

When creating an account with an email that already exists in the system, a user will now be redirected to the login page with a helpful message.

show_my_pr


app/controllers/users/registrations_controller.rb
```diff
@@ -4,7 +4,13 @@
 class Users::RegistrationsController < Devise::RegistrationsController
   # POST /resource
   def create
+    if User.where(:email => params[:user][:email]).any?
```
rlgreen91: Can you change the where clause to use the hash syntax?  Just to say `email: params[:user][:email]`

app/controllers/users/registrations_controller.rb
```diff
@@ -4,7 +4,13 @@
...
+      flash[:warning] = "Looks like that email is already taken you can login here or <a href=#{new_user_registration_path}>go back to sign up</a>".html_safe
+      redirect_to new_user_session_path(user: {email: params[:user][:email]}) and return
```
rlgreen91: Can you change the `and` to `&&` and add spaces inside the brackets (i.e., `{ email.... email] }`)?

Dfenniak: Hey @rlgreen91 because of how precedence works using `&&` here would break the functionality - there's a pretty good explanation of that here: https://stackoverflow.com/questions/39629976/ruby-return-vs-and-return

Dfenniak:
Thanks for the speedy review @rlgreen91 :) I've made the suggested updates except for one, and left a comment

rlgreen91:
Hi @Dfenniak, are you going to be able to finish this PR soon?

Dfenniak:
@rlgreen91 hey - so I updated the rubocop as per your request - as for the line break if you can tell me your preferred format I can do that. I was thinking of using `%Q` but then remembered it would insert line break chars :/

Let me know?

Dfenniak:
Also resolved merge conflicts

rlgreen91:
@Dfenniak thanks for making those changes!  For the line length one - one thing we've been doing more of is putting those messages in locale files and then referencing them when needed.  I would try that - you can look at some of our mailers to see how that would look.

Dfenniak:
closing this - as I didn't have time to keep iterating on it. sorry.

PR #3323, opened 2019-10-09
feat(#3313): Create location value object

### LINKED ISSUE #3313: Create location value object
Create an object `Location` that is a value object like `DateRange` (located in `app/models`).  The `Location` object should take in a required parameter state and optional parameters street location, city, and zipcode.  The street location should encompass house number, street name and type, apartment number, and direction (e.g., 124 West Avenue Suite #350 NW would be the street location).

The object should also define a `to_s` method that returns the concatenated parameters in order as well as a `==` method which returns true if all of the parameters have the same value in both objects.

sathia27:
Create location value object which resolve #3313

In your PR did you:


app/models/location.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def initialize(state:, street_location: nil, city: nil, zipcode: nil)
+    @state = state
+    @street_location = street_location
+    @city = city
+    @zipcode = zipcode
+  end
+
+  def to_s
+    [street_location, city, state, ("Zipcode: #{zipcode}" if zipcode)].compact.join(", ")
```
rlgreen91: Why the logic for zipcode?

sathia27: @rlgreen91  I could directly append. Only reason I went with this approach after going through some address format in internet. But other thing I observed is, address format looks different with different country. Let's discuss about the format and then I will make changes accordingly.

sathia27: @rlgreen91  I have removed the logic of zipcode for now. Let me know if we change format for address?

rlgreen91:
Hi @sathia27, thanks for removing that label for zipcode - all of our cases are in the US, where zipcodes are numeric.  As a result, we typically don't label them since it's pretty obvious which part is the zipcode.

Can you make a couple of last changes?  Can you have the `==` method resemble the one in the date range class, where the argument is `other` and we disable the style cop?  That will fix the Rubocop errors.  Lastly, I want to make the `to_s` method return the most common forms - can you add an argument `letter-style` that, if it's true, will return the address in this format:

`street location`
`city`, `state` `zipcode`

That way we cover both cases, where it is one line for inline text and letter style elsewhere.

sathia27:
@rlgreen91 
I made those changes which was suggested by you.
Also made some few changes to refactor rspec (like remove `should`) as suggested by `http://www.betterspecs.org/#should`

Let me know if any other changes need to be done.

rlgreen91:
Looks great, thanks!  I'll go ahead and merge.

PR #3364, opened 2019-10-12
Fix first-time database setup

k-stewart:
https://github.com/EBWiki/EBWiki/commit/603d8af7356a38e4d1544bec85f93e4f5561791d introduced some git conflict artifacts to db/structure.sql, which broke `rake db:setup` (and, presumably, other db setup tasks.) 

Removing the conflict artifacts allows db setup to proceed.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -1980,10 +1980,5 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20190422003442'),
 ('20190815012327'),
 ('20190824175709'),
-('20190825000337');
-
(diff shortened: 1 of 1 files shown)


k-stewart:
Hmm, looks like Travis is failing because it doesn't have AWS credentials set up.  Not sure if I can do anything about that from here, but let me know if there is.

rlgreen91:
Hi @k-stewart - thanks for catching that!  You don't have to worry about CI - that's an issue with how Travis handles environment variables for forks

Issue #3293, opened 2019-10-08
Use Devise confirmation mailer


rlgreen91:
Remove the confirmation email action and template from UserNotifier.  Add a file `spec/mailers/devise/confirmation_spec.rb` and model the spec after the example included in this: [Ruby Thursday guide](https://rubythursday.com/system/episodes/pdfs/000/000/069/original/Ruby_Snack_30_Testing_Devise_Mailers.pdf?1464299925).  Add a mailer preview by following this guide: [mailer preview](https://ionrails.com/2014/12/21/using-actionmailerpreview-with-devise-mailers-in-rails-4-1/)

k-stewart:
Hi team, I'd like to take a crack at this.

k-stewart:
Looks like the model is UserMailer, not UserNotifier, does that sound right?  I see references to UserNotifier in the docs, but its methods are defined in app/mailers/user_mailer.rb.

k-stewart:
Ah, i see, it was changed in a157d8ec64379bc0b57b37f0df71d9e1a0e47eba.  I can update the docs while i'm in here.

k-stewart:
Hmm, it looks like the docs [used to have the accurate model name](https://github.com/EBWiki/EBWiki/commit/72af37e08575a18c7a304352979e316b40689cd2),  and that was undone.  I'll leave it alone, then.

rlgreen91:
Addressed in #3371

PR #3371, opened 2019-10-12
Use Devise to confirm user emails

k-stewart:
Per #3293, use Devise's `confirmable` module to confirm user emails.

* Remove our home-rolled `UserMailer.send_confirmation_email`
* Add :confirmable to User
* Add `unconfirmed_email` column to `users`, per Devise's demands
* Add tests on the confirmation email
* Preview the mailer at http://localhost:3000/rails/mailers/devise_mailer

### DIFF
--- a/app/views/devise/mailer/confirmation_instructions.html.erb
+++ b/app/views/devise/mailer/confirmation_instructions.html.erb
@@ -1,5 +1,5 @@
-<p>Welcome <%= @email %>!</p>
+<p><%= "#{I18n.t('greetings.welcome')} #{@email}" %>!</p>
 
-<p>You can confirm your account email through the link below:</p>
+<p><%= I18n.t('notifier.confirmation_message') %></p>
 
-<p><%= link_to 'Confirm my account', confirmation_url(@resource, confirmation_token: @token) %></p>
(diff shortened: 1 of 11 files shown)


rlgreen91:
Hi @k-stewart, please resolve these merge conflicts when you get a chance.

app/views/devise/mailer/confirmation_instructions.html.erb
```diff
@@ -1,5 +1,5 @@
-<p>Welcome <%= @email %>!</p>
+<p><%= "#{I18n.t('greetings.welcome')} #{@email}" %>!</p>
```
rlgreen91: Rather than listing their email, let's list the user's name.

db/migrate/20191012043517_add_unconfirmed_email_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddUnconfirmedEmailToUsers < ActiveRecord::Migration[5.0]
```
rlgreen91: We don't need to include this column as we are not using `reconfirmable`.  Please remove.

k-stewart:
Sounds good, @rlgreen91.  Would you like me to amend this commit, or add a new one with the requested changes?

k-stewart:
I'll create separate commits, and you can squash them if desired 🚀

Issue #3311, opened 2019-10-09
Create a remove_admin_email action


rlgreen91:
In `AdminMailer`, create a new action and view called `remove_admin_email` that will notify recipients that the corresponding user is no longer an admin.  Then update the `SendAdminEmail` service object to call `send_admin_email` if a user's admin attribute is changed.

carloslazo:
I can take this on. I assigned it to myself.

rlgreen91:
Addressed in #3372

PR #3372, opened 2019-10-12
feat: creates remove admin action for mailer

carloslazo:
In your PR did you:

PR refers to issue #3311 

Let me know if you want me to edit the subject.

`subject: "#{@removed_admin.name} is no longer an admin.")`

or the view message

`<p><%= link_to "#{@removed_admin.name}", url_for(@removed_admin)%> has been removed as a admin for EBWiki.</p>`

  - [x ] Include a description of the changes?
  - [x ] Mention the issue the PR addresses?


carloslazo:
Also does EBWiki have a slack? I would like to be in the loop.

app/services/send_admin_email.rb
```diff
@@ -5,14 +5,30 @@ class SendAdminEmail
   include Service
 
   def call(user:)
+    # returns an array [false, true] or [true, false]
     previous_admin_status = user.previous_changes['admin']
-    send_new_admin_email(user: user) if previous_admin_status[1]
+    send_new_admin_email(user: user)
```
rlgreen91: I think you might have left this one in by mistake

app/services/send_admin_email.rb
```diff
@@ -5,14 +5,30 @@ class SendAdminEmail
...
+
+    if previous_admin_status[1]
+      send_new_admin_email(user: user)
+    elsif previous_admin_status[1] == false
```
rlgreen91: Let's change that to `!previous_admin_status[1]` - I'm surprised rubocop didn't yell about that

rlgreen91:
> Also does EBWiki have a slack? I would like to be in the loop.

Yes, we do, here's the link: https://join.slack.com/t/ebwiki/shared_invite/enQtNzkzMzkwNjY5MTY4LWE5NzQyMWZjZDJlZmMzMzJmYmJjYzVhOTJmZjNkYTYzMzVkOTkwOWUxYTZkY2U4ZGI2MzdkMWM4OTY5NzliNjg

app/services/send_admin_email.rb
carloslazo: Ah, good catch. Thanks.

app/services/send_admin_email.rb
carloslazo: That's cleaner. Thanks.

PR #3378, opened 2019-10-13
#3310 Add a delete button to case

### LINKED ISSUE #3310: Add delete button to case
Add a delete button to the case show view that, when, clicked, will trigger the `destroy` action in the cases controller.  This button should only be visible to admins, and should be an orange button in the upper right hand corner.

danuluma:
In your PR did you:

Adds a delete button to the case show view that, when, clicked, triggers the destroy action in the cases controller. This button is only be visible to admins, and is an orange button in the upper right hand corner.
Fixes #3310 

Normal user(the delete button is not available)
Screenshot 2019-10-10 at 14 03 41

Admin user(can delete a case)
Screenshot 2019-10-10 at 14 02 37

After deletion
image


app/views/cases/show.html.erb
```diff
@@ -33,31 +33,38 @@
...
               <% end %>
-            <% else %>
-              <%= link_to 'Follow This Case', follows_case_path(@this_case), method: :post, class: "btn-lg btn-primary text-center", data: { toggle: 'popover', container: 'body', trigger: 'hover', placement: 'bottom', title: 'Login to Follow this case!', content: 'Click here to receive an email whenever this case is updated.' } %>
-            <% end %>
-          </div>
-      </h1>
+            </div>
+            
+        </h1>
+        <% if current_user && current_user.admin? %>
```
rlgreen91: @danuluma Can you change this conditional to match the way we display the edit profile button on the user show page?

danuluma: Sure, let me do that

rlgreen91: We did add a case policy so that should make it a bit easier.

rlgreen91:
Hi @danuluma, are you going to be able to finish this PR soon?

Issue #132, opened 2015-08-27
Need to clarify EBWiki branding


campgurus:
from feedback: what does the "EB" stand for? The abbreviated name is memorable; I found myself wondering what it stood for though -- maybe make that more prominent somewhere on the home page? it doesn't even appear on the About page, FWIW

campgurus:
A slogan/tagline may solve this issue.

rlgreen91:
At this point, on the about page, start the first paragraph off with "End Bias Wiki (EBWiki)" to establish what the name means.

rlgreen91:
Addressed in #3438

PR #3438, opened 2019-10-18
Add org's full name to about page

k-stewart:
[#132]


k-stewart:
I've been taking some small tickets without waiting for approval, i hope that's ok.  Feel free to ignore if it's not helpful.

PR #3470, opened 2019-10-30
Deleted lines of code that display cases with no avatar

daniellecroxton:
In regards to issue: On the home page, only use cases with pictures in the carousel #3013 

I removed the `else` statement for the home page thumbnails, leaving the page to only show the cases `if` an avatar is present.

I struggled a bit with adding an avatar to the seed data to get a screenshot showing the home page with one avatar. However, I did confirm that with this change none of the seed data shows, since none have avatars.

In your PR did you:

Screen Shot 2019-10-29 at 8 59 17 PM


daniellecroxton:
Hi, I just realized that the partial that I edited to adjust which cases show on the home page (only those with avatars) is also used on: 

1. The user's profile page
2. The search results page
3. The agencies page

I assume we still want cases without avatars to show in these instances, so I propose editing the home page to not use the partial and leaving the partial in tact for the 3 instances noted above. I will commit these changes, but if you disagree, please let me know!

rlgreen91:
@daniellecroxton thanks, for pointing that out.  I originally wrote this as a temporary solution, but you're right, this affects other pages as well.  We did start the process for having a default avatar for all cases - if you add one right now you'll see that an avatar is currently created (it's just the placeholder picture).  But, we didn't finish the rest of the changes needed to verify that the default avatar is always attached.  Do you want to take that on?  If so, let me know and I can write a better issue for that and assign it to you.

daniellecroxton:
@rlgreen91 Yes, sounds good! Thank you.

rlgreen91:
Ok, so I'll close this and that issue for now.

PR #3498, opened 2019-11-20
Remove :try and :blank? checks from cases controller

k-stewart:
They are unnecessary, since an empty ActiveRecord array responds to `:order` and the code `rescue`s if it can't find the case.

@rlgreen91 I'd like to add a controller spec for this, just to assert that the history is populated correctly in both cases. I see that the repo doesn't have any controller specs at present. Could you give me some context around that?

[#3197]


rlgreen91:
@k-stewart Yes, so we're following Rails in general by moving away from controller specs and towards request specs.  We have a request spec that covers the `history` action currently.  If you want to flesh that spec out a bit more, then feel free to do so, but otherwise, this looks good for me to merge.

k-stewart:
That makes sense. Occasionally I feel some friction between the relatively high cost of request specs and the desire to thoroughly test actions, but I will work that out internally : )

PR #3512, opened 2019-11-24
3144 remove ahoy

trystant:
In your PR did you:

  - [X] Include a description of the changes? I removed all traces of ahoy from the codebase with the exception of the tables, since they still have data that can be used & incorporated into a new package.


rlgreen91:
@trystant I understand removing the Ahoy related code, but why do we need to remove all analytics code?

trystant:
@rlgreen91 I want to remove this analytics code from `master` because it's not being used currently. An analytics solution with either SumoLogic or MixPanel will need to be readjusted. When that solution works, we can use `git` history to bring the analytics specs back and then create a clean implementation.

rlgreen91:
@trystant Ok, thanks.  Once the tests pass again I'll approve so you can merge.

trystant:
@rlgreen91 FWIW, I also want to remove the Maps code as well until a new(er) implementation is in place.

rlgreen91:
> @rlgreen91 FWIW, I also want to remove the Maps code as well until a new(er) implementation is in place.

That sounds good, just put it in a separate PR please.

PR #3532, opened 2019-12-01
Ruby 2 6 upgrade

rlgreen91:
Upgrades Ruby version to 2.6.5


trystant:
@rlgreen91 I think that you need to update the version of rubocop used in codeclimate because it's different from the version in the rubocop gem by default.

PR #3604, opened 2020-06-19
3571 remove versions assignment and undo

trystant:
In your PR did you:

  - [X] Include a description of the changes? Removes the unused undo action and route. Also makes a slight change to the versions_controller request spec


trystant:
> Could we remove this spec altogether for right now @trystant ?

Why remove this spec? Were we getting rid of undo altogether?

rlgreen91:
> > Could we remove this spec altogether for right now @trystant ?
> 
> Why remove this spec? Were we getting rid of undo altogether?

No, but given that the spec covers a feature that doesn't currently exist and won't exist in that form in the future, I'm not sure I see a reason to keep it around in its current form.

PR #3617, opened 2020-07-08
Upgrade gems for 51

rlgreen91:
In your PR did you:


trystant:
While the Gemfile* upgrades are good, isn't the deletion of the `doc/app` directory something separate? Is that an essential part of this upgrade?

rlgreen91:
@trystant that directory was only associated with sdoc, which I removed, so there's not really a need to keep it

PR #3689, opened 2020-09-30
Improvements to Documentation and default .env Setup

iarobinson:
### Description:

As I was getting this project running locally, I ran into a few issues which slowed me down. This PR is me adjusting the documentation and the sample files so that future developers can get started a little bit quicker.

#### Improvments:
- Upgrade the .env.example file to have a default server set so new developers don't need to specify a local for elastic search
Currently, the developer needs to adjust the elastic search URL. I just made it so the new developer doesn't need to set it.
- Upgrade README.md with information on how to set up Elasticsearch
If a new developer has never used elastic search before, this could be a point where we loose them.

This PR does not address an issue. It's just what I noticed while getting the app running locally myself.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,6 @@ AWS_SECRET_KEY_ID=Get credentials from Amazon aws
 MAILCHIMP_API_KEY=Get this from Mailchimp or talk to someone at EBWiki
 MAILCHIMP_LINK=Get this from Mailchimp or talk to someone at EBWiki
 MAILCHIMP_LIST_ID=Get this from Mailchimp or talk to someone at EBWiki
-SEARCHBOX_URL=Your friendly neighborhood Elasticsearch server
+SEARCHBOX_URL=http://localhost:9200
 CODECLIMATE_REPO_TOKEN=EBWiki devs can provide a token if needed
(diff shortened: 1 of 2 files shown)


rlgreen91:
@iarobinson Hey, this looks good.  Can you resolve the merge conflicts and re-push?

managedkaos:
I think the main cause of the conflicts is that the edits to `docs/SETUP_LOCALLY.md` should now go into `docs/SETUP_LOCALLY_FULL_STACK.md`.

I can help out if you'd like.

rlgreen91:
@iarobinson Wanted to follow up on this - just need to resolve the merge conflicts.

rlgreen91:
I wanted to follow up on this: looking at this now, I'm not completely sure we need these changes.  We have updated `.env.example` in a separate PR to include the default Elasticsearch URL.  However, in terms of instructions on how to set up Elasticsearch, we already point readers to their guides on how to set it up for their environments.  Additionally, this is now the alternative method to running EBWiki locally - we try to steer people towards using the Docker setup instead, which will take care of this for them.  Because of that, I'm going to close this.

PR #3729, opened 2020-10-02
Add cause of death name on case show view page

### LINKED ISSUE #3681: Display cause of death on case show view
Update the case show view to display the `cause_of_death_name` attribute as "Cause of death: [value]".  This should be below the list of agencies, with the same font size.  You'll want to display the values in the locale file instead of the actual enum, similar to how we do the dropdown field in the case form.

93lykevin:
Add cause of death name on case show view page
closes #3681 
In your PR did you:


93lykevin:
I'm not sure what all these failed tests are, but I think I've got it. Please review it and let me know if there is anything you'd like me to fix :)

rlgreen91:
Hey @93lykevin, that's related to another thing, I wouldn't worry about that.  This looks good in general.

rlgreen91 (approved):

PR #3792, opened 2020-10-09
Selection sort

Prajesh-Srivastava:
Selection sort is an algorithm that splits a list of elements into sorted and unsorted. During each iteration, the algorithm will traverse through the unsorted list and add the smallest element to the end of the sorted list.

It adds to the sorted group by swapping the smallest number in the unsorted group with the first element in the unsorted group.

In your PR did you:

### DIFF
--- a/Selection sort
+++ b/Selection sort
@@ -1,0 +1,26 @@
+def selection_sort(lst):
+    """
+    Selection sort function
+    :param lst: List of integers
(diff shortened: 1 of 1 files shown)


rlgreen91:
Was this someone submitted here by mistake?  Not only does it not address any issues we have, it's not even the right language.

Issue #3841, opened 2020-10-24
Remove rollbar from development environment


trystant:
After having years of data, we don't need to determine how many errors started in development. In comparison, the noise level is very high. Removing rollbar calls from development will help reduce the noise.

PR #3842, opened 2020-10-24
Added dev environment to those where Rollbar isn't called #3841

trystant:
In your PR did you:

  - [X] Include a description of the changes?  Disabled rollbar in development.

### DIFF
--- a/config/initializers/rollbar.rb
+++ b/config/initializers/rollbar.rb
@@ -11,7 +11,7 @@ Rollbar.configure do |config|
   config.environment = ENV['ROLLBAR_ENV']
 
   # Here we'll disable in 'test':
-  config.enabled = false if Rails.env.test?
+  config.enabled = false if Rails.env.test? || Rails.env.development?
 
   config.exception_level_filters['ActionController::RoutingError'] = 'ignore'
 


config/initializers/rollbar.rb
```diff
@@ -10,8 +10,8 @@
 
   config.environment = ENV['ROLLBAR_ENV']
 
-  # Here we'll disable in 'test':
-  config.enabled = false if Rails.env.test?
+  # Here we'll disable in 'test' and 'development':
+  config.enabled = false if Rails.env.test? or Rails.env.development?
```
mockdeep, line 14: @trystant `or` [has super weird operator precedence](https://stackoverflow.com/questions/2083112/difference-between-or-and-in-ruby), so I'd generally go with `||` in situations like this. That being said, [I tried a similar change](https://github.com/EBWiki/EBWiki/pull/3883) before I saw this PR, but it sounds like we might want to discuss what approach to take with this.

mnyon-grandkru: @mockdeep Yeah, `or` is lower on the [Ruby operator precedence totem pole](https://stackoverflow.com/questions/21060234/ruby-operator-precedence-table). I'll change that up.

rlgreen91 (approved):

PR #3874, opened 2021-01-17
Rubocop: install specific rubocop version for CI

mockdeep:
**What**

This updates the CI configuration for Rubocop to install the same
version that is specified in `Gemfile.lock`.

**Why**

This should get the build passing. It's a quick fix. I think it might be
better to run `bundle` instead of installing `rubocop` directly. Caching
should reduce performance overhead.


mockdeep:
Closing in favor of https://github.com/EBWiki/EBWiki/pull/3875.

PR #3875, opened 2021-01-18
CI: specify rubocop version for build

mockdeep:
**What**

Update the CI configuration for Rubocop to install the same version that
is specified in `Gemfile.lock`. I think it might be better to run
`bundle` instead of installing `rubocop` directly, but it will be good
to set up caching to reduce performance overhead.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -13,7 +13,7 @@ jobs:
         run: |
           PATH="$(ruby -r rubygems -e 'puts Gem.user_dir')/bin:$PATH"
           echo $PATH
-          gem install --user-install rubocop
+          gem install --user-install rubocop -v=0.93.1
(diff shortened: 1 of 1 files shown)


mockdeep:
Woot! Passing build. I think I'll probably break this out into separate PRs.

trystant:
Thanks for this @mockdeep. It does make sense to split this into separate PRs. There may be issues that already exist for some of the fixes here as well.

mockdeep:
@trystant sorry about that, it's passing again. The first 2 PRs in the chain are #3884 and #3885. They'll need to be merged with a broken build until we can get everything merged.

mockdeep:
Okie dokey, this is the last step for a passing build!

managedkaos (approved):
LGTM

mockdeep:
After this, we should be able to lock down the builds again in order to prevent breakages from being introduced.

PR #3844, opened 2020-10-24
Replaced access_token with ROLLBAR_ACCESS_TOKEN #3843

trystant:
In your PR did you:

  - [X] Include a description of the changes? Removes the access token from the initalizer and relies on ROLLBAR_ACCESS_TOKEN
  - [X] Mention the issue the PR addresses?
  - [ ] Include screenshots of any changes to the UI?
  - [ ] Isolate any changes to gems (meaning that any new, updated, or removed gems and resulting code changes should be in their own PR)?
  - [ ] Add and/or update specs for your code?

PR #3883, opened 2021-01-18
CI: disable rollbar in dev mode

mockdeep:
**What**

This makes it so that Rollbar only runs in staging and production mode.

**Notes**

Rollbar was already disabled for test mode, but errors were triggered on
CI, which makes me think the tests are running in dev mode instead of
test.

### DIFF
--- a/config/initializers/rollbar.rb
+++ b/config/initializers/rollbar.rb
@@ -10,8 +10,7 @@ Rollbar.configure do |config|
 
   config.environment = ENV['ROLLBAR_ENV']
 
-  # Here we'll disable in 'test':
-  config.enabled = false if Rails.env.test?
+  config.enabled = Rails.env.staging? || Rails.env.production?
 
   config.exception_level_filters['ActionController::RoutingError'] = 'ignore'
 


mockdeep:
This won't pass the build, but it's a step towards a passing build without Rollbar errors popping up in the meantime.

mockdeep:
@rlgreen91 @trystant @managedkaos not sure who has access to admin settings for the repo, but can we disable the required builds until this series of PRs is merged?

rlgreen91:
Thanks @mockdeep, but we do not want to disable Rollbar for dev mode as it helps us debug unexpected issues, particularly when working with contributors who are new to development and have little experience with debugging errors.

mockdeep:
@rlgreen91 it looks like this creates a lot of noise in the issues. I cause a lot of errors when I'm developing, and wouldn't want to cause a flood of issues here. I wonder if it could be set up so that only production/staging create issues on Github? Also, I'm wondering how it'll work with #3844. Do you plan to give the key to new people on the project?

PR #3884, opened 2021-01-18
CI: fix Docker setup

mockdeep:
**What**

* Downcase `PWD`. I'm guessing these files were developed on a file
  system that is not case sensitive. Linux, on the other hand, is.
* Update `Makefile` to use [the `shell` keyword][shell]. `pwd` was
  returning nothing, so it wasn't able to properly find the config file
  for the `codeclimate` task.

[shell]: https://stackoverflow.com/a/18137056/372479


rlgreen91 (approved):
LGTM

rlgreen91:
This LGTM @mockdeep but you might want to get the ok from either @trystant or @managedkaos before merging.

mockdeep:
ping @trystant, @managedkaos. This and [my rubocop fix](https://github.com/EBWiki/EBWiki/pull/3875) should get the build passing on `main`.

managedkaos:
regarding `$(PWD)` case insensitivity, the case you're referring to is a Makefile vs Bash construct.  Makefile reads `$(X)` as the environment variable defined by $X.  Bash on the other hand reads $(X) as the value returned by the command X.

So the changes work but I just want to comment on why they work where they work. 👍

managedkaos (approved):

mockdeep:
@managedkaos my machine (and CI) complained that there was no such command `PWD`. When I downcased them, it stopped complaining. But then a new issue surfaced after that, where I had to add the `shell` command. What you say explains the second bit of my commit, where `$(pwd)` was coming back with nothing.

PR #3797, opened 2020-10-10
Fixes #3796 upgrade rails_admin gem

Morrism1:
In your PR did you:

  - Upgraded Rails_admin gem to ~> 2.0
  - Fixes #3796 upgrade rails_admin gem

- [x] Include a description of the changes?
- [x] Mention the issue the PR addresses?
- [x] Isolate any changes to gems (meaning that any new, updated, or removed gems and resulting code changes should be in their own PR)?
- [x] - Add and/or update specs for your code?

PR #3885, opened 2021-01-18
CI: fix Sprockets issues

mockdeep:
**What**

* Introduce `app/assets/config/manifest.js`. Sprockets was [recently
  upgraded][upgrade] to version 4, but [it now requires this file][file]
  to exist in order to determine which files to compile.
* Disable concurrent exports for Sprockets. It has a tendency to [cause
  segfaults][segfault] in Ruby, making builds (and deploys) super flaky.
  I've already seen it fail a few times working on getting the build
  passing.

[upgrade]: https://github.com/EBWiki/EBWiki/pull/3797
[file]: https://github.com/rails/sprockets/blob/master/UPGRADING.md#manifestjs
[segfault]: https://github.com/sass/sassc-ruby/issues/207

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,0 +1,4 @@
+//= link_tree ../images
+//= link application.css
+//= link application.js
+//= link favicon/browserconfig.xml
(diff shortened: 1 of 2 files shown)


rlgreen91 (commented):
@mockdeep Have you confirmed this doesn't cause problems with our current version of Bootstrap?  We're still on 3.  If you can confirm that, then I think this is good to go.

mockdeep:
@rlgreen91 I looked it over and didn't see any differences between Sprockets 3 and 4 visually. I don't think it should impact any dependencies unless they rely on Sprockets directly, which Bootstrap doesn't.

Before this and any other PRs can be merged, we'll need to relax the branch protection rules until we can get the build passing.

rlgreen91 (approved):

PR #3886, opened 2021-01-24
Adds release task to automatically run migrations during Heroku deployment

rlgreen91:
Adds a release task Bash script to run a set of rake tasks during deployments to staging and production on Heroku.  This script will be run after the code has been pushed to the environment and prior to actual deployment of the app.

Currently, the script only includes a command to run migrations.  The PR template has been updated with a reminder to verify that the script has been updated if needed.  Also, the isolated gem item has been removed from the PR template; gem changes are usually in their own issue, so there was usually no need for this check.


release-tasks.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
mockdeep, line 1: I've been seeing more things moving towards `/usr/bin/env bash`. Probably doesn't matter in this case, but that [respects any local overrides](https://stackoverflow.com/a/2429517/372479) in `PATH`.

```suggestion
#!/usr/bin/env bash
```

release-tasks.sh
```diff
@@ -0,0 +1,4 @@
...
+
+echo "*** Running any new migrations ***"
+rake db:migrate
```
mockdeep, line 4: I don't remember which version it was introduced, but `rails` [is now the default command](https://guides.rubyonrails.org/command_line.html) instead of using `rake`. They do the same thing in this case, but I've been slowly adapting to the new way. I'd probably also add a `bundle exec` to prevent any environment weirdness:

```suggestion
bundle exec rails db:migrate
```

Procfile
```diff
@@ -1 +1,2 @@
+release: ./release-tasks.sh
```
mockdeep, line 1: It looks like this file isn't executable, so I think this might throw an error. You should be able to `chmod +x release-tasks.sh` and commit the permissions change.

mockdeep (commented):
I love that this task exists. I recently started to use it on my projects, too!

release-tasks.sh
rlgreen91: Hmm, actually, I decided to go with rake over rails because I wanted to avoid any issues with this being a bash executable.  But if rails will work fine then I'll switch it.

managedkaos (changes requested):
Looks good to me!  The only question I would have is "does it work in staging?" :)

Also, is it possible to share the output of a test run?

release-tasks.sh
mockdeep: @rlgreen91 I can't think of a scenario where this would work with `rake` but `rails` would not be available. The `db:migrate` task depends on the entire Rails environment being loaded, so it won't work unless Rails is present.

That being said, I don't think there's any practical difference, so no biggie if you want to leave it as `rake`.

trystant: I'd say let's go with `rails` if it works now.

trystant:
> Looks good to me! The only question I would have is "does it work in staging?" :)
> 
> Also, is it possible to share the output of a test run?

I agree with @managedkaos here

rlgreen91:
@trystant @managedkaos  Well, I've made those changes above so when I push up and merge, we will see if it runs in staging :)

What did you have in mind in terms of the output of a test run?  The results should show up in the deployment logs on Heroku

Procfile
mockdeep: @rlgreen91 it looks like you didn't make `release-tasks.sh` executable. I think the next time you deploy it will throw an error. Just FYI.

managedkaos:
Ok that works. I can't see heroku though... 🤔🤷🏾‍♂️

mnyon-grandkru:
I've added @managedkaos to the list of users in Heroku and I think that the output of the test deploy would suffice.

PR #3890, opened 2021-01-31
Security: use secure git source in Gemfile

mockdeep:
When you run `bundle` you'll see a warning about an insecure source. The
`git://` protocol does not use `https`, so it is vulnerable to
man-in-the-middle attacks. Bundler [provides a `git_source`
helper][helper] where we can instead provide our own secure URL, so this
switches `Gemfile` to use that.

[helper]: https://bundler.io/guides/git.html


rlgreen91 (approved):
LGTM

PR #3891, opened 2021-01-31
Upgrades: update Ruby to 2.6.6

mockdeep:
I've been seeing some internal Ruby errors when running tests locally,
so this upgrades to the latest patch release in hopes of fixing them.
Ruby 2.7 is the latest, but that'll be a harder upgrade, so for now
sticking to 2.6.6.


managedkaos (commented):
LTGM.  I will defer to @rlgreen91 and @trystant though since this is more of an application environment thing.  

The Dockerfile looks good, though, and the checks are passing.

Down the road we can remove the `.travis.yml` file altogether since we are off TravisCI now.

rlgreen91 (approved):

rlgreen91:
We're not upgrading to Ruby 2.7 for a while, as we need to get our gems and Rails upgraded first to catch up.  Otherwise this is fine to merge.

PR #3897, opened 2021-02-01
Coverage: write tests for CaseSearch

mockdeep:
This adds tests for the `CaseSearch` class. I'm not sure why, but I
didn't need to add `search: true` to enable `Searchkick` callbacks. I
get the same test result either way, so I decided to leave it off [for
the sake of performance][perf].

I also needed to configure `Webmock` to allow connecting to `localhost`
in order to connect to `elasticsearch`. I pulled it out into its own
file even though it's only a couple of lines. `rails_helper.rb` can
become a bit of a junk drawer of configurations for a wide variety of
related gems, so I find it's nice to have each thing configured
separately.

[perf]: https://github.com/ankane/searchkick#testing


rlgreen91 (changes requested):
While this looks good overall, there are failing specs in the CI run that need to be addressed.

rlgreen91:
@mockdeep It still looks like there are still some issues with the specs you added.  We're going to upgrade the Rails version, so I'm going to close this.  If possible, I'll try to incorporate your changes into the code as part of the upgrade or if you'd like to do it yourself later just reply and let me know.

PR #3898, opened 2021-02-01
Upgrades: remove deprecated configuration

mockdeep:
This removes the setting `raise_in_transactional_callbacks`, as it no
longer has any effect and is removed in Rails 5.1.


rlgreen91 (approved):
LGTM

PR #3899, opened 2021-02-01
CI: try removing call to `prewarm` script

mockdeep:
The CI build seems to be pretty slow and flaky. Most of the time when I
see it fail it is due to the `prewarm` script timing out. I'm wondering
if removing the call to the script will speed things up at all.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,6 @@ build:
 run:
 	docker run --detach --rm --volume "$(shell pwd)":/usr/src/ebwiki \
 		--publish 3000:3000 --name ebwiki ebwiki/ebwiki
-	./dev_provisions/prewarm.sh
 
 updatedb:
 	docker exec -e PGPASSWORD=ebwiki ebwiki pg_restore --verbose --clean \


managedkaos:
@mockdeep if i recall, the prewarm script is used to get the site and database ready before any requests come in.  without it, the first request can either take a really long time or fail.  So at best the first request will take a while; at worst it will fail.

Try this:

1. open two terminals and move to the project dir
2. in terminal 1, start the server without detaching: `docker run --rm --volume "$(pwd)":/usr/src/ebwiki --publish 3000:3000 --name ebwiki ebwiki/ebwiki`
3. wait for the server to be ready
4. in terminal 2, curl the site and see how long it takes: `curl -s -o /dev/null -w "%{time_total}" "localhost:3000"`
5. now rerun the experiment with prewarming.
6. terminal 1: stop the server, start it again, run `prewarm.sh`
7. terminal 2: `curl -s -o /dev/null -w "%{time_total}" "localhost:3000"`
8. compare the times from steps 4 and 7 and consider the prewarm time.

On my system, the first request took `6.470288s` without warming up.  The second request took 14 seconds to warm up and `1.994077s` for the first request.

I think waiting ~16s for a good first request isn't too long, especially if it's automated and confidence is high that the request will be good.

Also, I think the cases where the prewarm timed out during CI have been resolved by your recent changes.  Can you share one/two?

I don't have a record of requests that failed without warming up but I'll see if I can recall/find any.  But I think they were all rspec tests.

As it is, I think the `rspec` workflow is the only one that needs the docker container to be built and running.  Can you think of another way to run specs that need the server without the server and/or database?  Maybe we can isolate the specs that need the complete environment and put them in a separate workflow?

managedkaos:
ahhh I'm also recalling that if tests run _immediately_ after the container is started they will indeed fail.

Try this:
```
docker run --detach --rm --volume "$(pwd)":/usr/src/ebwiki --publish 3000:3000 --name ebwiki ebwiki/ebwiki && curl -s -o /dev/null -w "%{time_total}  %{http_code}" "localhost:3000"
deda2437f21a34139aad9cee6bbbac8ea06e4af7280b52f418df609821cbd28f
```

In my case the curl fails immediately with code `000`.  

So if any tests are to run after the container starts, we need to wait until the server is actually ready to receive requests.  That's where the prewarm script comes in because it keeps anything else from happening until the server is ready.

Again, I think the timeouts were fixed by your previous PRs but let me know if anything was missed.

managedkaos:
one last comment...  😅 

You could also try replacing the prewarm script with a sleep statement but I think the prewarm is faster and more reliable.  The startup time is nondeterministic since it depends on the system where the container is running.  so a fixed sleep time could be too short and cause requests to fail if the sleep cycle wasn't long enough.  or it could be too long causing wasted cycles.

As for the prewarm script, it keeps checking the site every 5 or 6 seconds until its ready.  so it will wait the perfect amount of time for the  system where the container is running.

managedkaos:
I lied.  one more thing :) 

workflows that need to run the site in the container may include updates to gems.  in that case, the image needs to be built as a part of the workflow.  I'll admit, that adds time to the workflow.

the alternative would be to pull down a prebuilt image and use that but in the case of updated gems or some other changes, the workflow may fail if the pulled image is out of date with the newly updated code.

mockdeep:
@managedkaos here are a couple of recent builds that failed due to what looks like prewarm timeouts: [1 build](https://github.com/EBWiki/EBWiki/runs/1724553867?check_suite_focus=true), [2 build](https://github.com/EBWiki/EBWiki/runs/1801889441?check_suite_focus=true).

I'm not very familiar with Docker at all. I've tried it several times and failed to get it working for myself. This PR was mostly a shot in the dark to see what impact it might have, but it's failing RSpec due to the database not being ready. Still, 7-8 minutes seems like an awfully long time for a CI build on such a small app. I think there are probably a few things that contribute to the problem.

* `DatabaseCleaner` can be super slow. It should really be configured for transactions on tests that don't use the browser. It'll be a non-issue, though, once we upgrade to Rails 5.1, where they've set up tests to use shared database connections. At that point we can remove DatabaseCleaner and rely on the builtin setup, using transactions for *all* tests.
* I think Github Actions is already running inside of a container (`ubuntu-latest`) and then we're pulling down a Docker image on top of that to run the specs in. It seems like we are adding some overhead to the process by nesting another image.
* It doesn't look like we're taking advantage of any caching on CI. I'm not entirely sure what is the best way to go about it. I've done a lot of CI work without docker where you can take advantage of the CI builtin caching for things like gems and compiled assets. Inside of a Docker image, though, I think we'd need to do some extra work to copy cached assets in and/or be regularly pushing a new image with the latest dependencies every time there is a gem upgrade.

I do think there's value in having Docker for easier contributor setup, especially people less familiar with Rails. But it seems like it requires very active development to keep it up to date, so I'm not sure what to do about that. Maybe we could set up a Github Action to automatically build and push a new image every time gems are updated.

Side note, in [this wiki page](https://github.com/EBWiki/EBWiki/blob/main/docs/SETUP_LOCALLY.md#set-up) it says 3-5 minutes for the server to boot up. Is that a typo? Otherwise, that seems like a pretty long time to wait.

managedkaos:
> it's failing RSpec due to the database not being ready

Yep, that's part of the reason we need to wait for everything to be ready.  When I say "server" i mean the monolithic application that is running the webapp and the database.

> 7-8 minutes seems like an awfully long time for a CI build on such a small app

Its no so bad, in my opinion.  Keep in mind that we are getting free compute on shared services from GitHub.  If we could provision our own instances we could use bigger, faster servers.  And again, most of the build time is for the docker image.  Speaking of...

> we are adding some overhead to the process by nesting another image

I think you'll find that most modern, publicly available build systems operate this way.  Bitbucket Pipelines, Azure DevOps Pipelines, GitLab CI, etc.  To get a bare metal server for CI you would likely need a self-hosted build system like Jenkins, TeamCity, Bamboo, or GoCD.

> It doesn't look like we're taking advantage of any caching on CI. I'm not entirely sure what is the best way to go about it.

We are not caching and I'm not sure how to go about it either.  I'm open to suggestions though!  GitHub Actions has options for caching libraries and such at the code level, like if we were running Rails inside the existing container.  However, our rails stuff is going on inside the container we build.  There is also docker image layer caching, but again, since we need to (that is, must) make sure the gems are updated on each build layer caching might not help.  Again, I'm open to suggestions.

> I do think there's value in having Docker for easier contributor setup, especially people less familiar with Rails.

I agree! 😄  I'm one of those people that is not very familiar with rails.  Having the container really got me up to speed.

> Maybe we could set up a Github Action to automatically build and push a new image every time gems are updated.

We build and push a new image with every merge to the master branch.  [This workflow does it](https://github.com/EBWiki/EBWiki/actions?query=workflow%3A%22Publish+Docker+Image%22).

> this wiki page it says 3-5 minutes for the server to boot up. Is that a typo?

Nope, not a typo.  Specifically it says to "boot up" not build.  In that section of the document, the reader/user would be pulling a prenbuilt image and then starting it.  With no build time, the server should boot in the described time.

managedkaos:
As I was leaving the last comment, I was brainstorming the discussion around running our environment in a container _on_ a container.  We may be able to take advantage of caching and a faster, more reliable build and execution if we run the services inside the container where the job is running.  That is, use the Actions container itself.

I did a quick search and it looks like that approach might work.

I'm going to close this PR and create an issue with the links to the docs I found.  If you don't mind, I'll assign it to you and we can review what you come up with.  Should be pretty cool! 👍

PR #3902, opened 2021-02-03
Deploy: make release task executable

mockdeep:
This updates the permissions on `release-tasks.sh` to allow executing it
directly.


rlgreen91:
The release task runs, so there's not really a need for this at this point.

PR #3904, opened 2021-02-03
Bugs: fix url redirects

mockdeep:
**What**

There is an issue in production where sometimes a user will be
redirected to an odd url with a random port, something like
`ebwiki.org:9230/agencies/1`. It's not immediately clear why this is
happening, but one thing that we noticed is that the
`ApplicationController#default_url_options` is *always* setting a port
for the generated URL, regardless of environment. This might be
interfering with the browser's native port management in some way, so
this PR moves the setting to the environment configurations, making it
so that we only set the port for dev/test mode, but leave the port off
for production and staging.

**Notes**

This also adds a `protocol:` parameter for `staging` and `production` so
that URLs are generated with `https://`, avoiding an extra redirect from
`http` to `https`.

We probably also want to change the path helpers used to redirect in the
controllers from `agencies_url` to `agencies_path`. Practically
speaking, this should not matter, but the first produces the entire url
(`https://ebwiki.org/agencies`) whereas the second produces just the
path (`/agencies`).


rlgreen91 (approved):
LGTM

rlgreen91:
> We probably also want to change the path helpers used to redirect in the
> controllers from agencies_url to agencies_path. Practically
> speaking, this should not matter, but the first produces the entire url
> (https://ebwiki.org/agencies) whereas the second produces just the
> path (/agencies).

Yeah, I did note that as a to-do for later.  Once we confirm this addresses that redirect issue, I'll create tickets for that - I doubt this is the only controller we have that pattern in.

PR #3907, opened 2021-02-06
A hotfix for missing end in carrierwave code

rlgreen91:
Quite literally a hotfix for a missing end statement.  I'll let it go through CI but otherwise merge immediately.


mockdeep:
Should we re-enable the branch protection rules?

PR #3919, opened 2021-02-07
Fixes seeds file and downgrades sprockets

rlgreen91:
Fixes the seeds file and removes the production database backup.  Updates the docs accordingly.  Also downgrades Sprockets due to seg fault issues.

More detail: Since the seeds file is in better shape, I went ahead and removed the production database backup.  We can find another spot for that, but I wasn't a huge fan of having that in the repo, since we could potentially expose personal information to anyone that clones the repo.  To that end, I included a note that if someone does need something closer to the prod database to work on a ticket to let us know.

I also downgraded Sprockets from 4 to 3.7.2 - Sprockets 4 is causing consistent segfaults when run on Linux distros with a kernel version of 5+, [as described here](https://github.com/rails/sprockets/issues/633).  Once that issue is addressed, we'll be able to just remove the line specifying the version in the Gemfile; everything else is still configured to work for 4.

In your PR did you:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -168,3 +168,5 @@ gem 'recaptcha'
 group :production do
   gem 'cloudflare-rails', '~> 0.6'
 end
+
+gem 'sprockets', '3.7.2'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -735,7 +735,7 @@ GEM
       builder (~> 3.0)
     social-share-button (1.2.3)
       coffee-rails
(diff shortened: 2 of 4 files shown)


mockdeep:
@rlgreen91 [my PR here](https://github.com/EBWiki/EBWiki/pull/3926) should fix sprockets for you. Can you test it out?

mnyon-grandkru (approved):
LGTM

PR #3926, opened 2021-02-07
Dev: nest export_concurrent config in block

mockdeep:
I'm not sure what the difference is, but when it's the single line
format I still see a bunch of concurrency errors when compiling assets,
but when I put it in the block the error stops happening.


rlgreen91:
Right now Sprockets is intentionally set to 3.7.2, as 4 ran into segfault issues on some Linux versions.  I'm guessing this rspec error is because this isn't a feature of Sprockets 3 - when I get a chance, I'll pull this branch down and confirm that this fixes the issue with 4 on Linux kernels that are 5+.

rlgreen91 (approved):
Thanks

Issue #3389, opened 2019-10-15
Add location object to Agency and Organization models


rlgreen91:
Add the location object to the `agency` and `organization` models.  You can use the date range and address examples in [this guide](https://revs.runtime-revolution.com/value-objects-in-ruby-on-rails-9df64bc8db34) as examples of how the model code should look.

jm96441n:
Does this still need to be addressed? I'm interested in contributing and thought this looked like a good issue to start on

rlgreen91:
Hi @jm96441n - sorry for the delayed response!  Yes, feel free to pick it up if you'd like.

Xeitor:
Hello, can I work on this issue?

danielvdao:
Hi @rlgreen91, I think I may have tried a pass at this. Please let me know if it's not relevant anymore: https://github.com/EBWiki/EBWiki/pull/3973. I looked at this a week or two ago and this was one of the first open issues I saw. Turns out there's more now 🙂 .

rlgreen91:
Thanks @danielvdao - I took a look just now.

rlgreen91:
Merged

PR #3973, opened 2021-02-26
Add Location value object to Agency model

danielvdao:
### Description

Hi, I heard about this project and was interested in working on some of the issues y'all had and trying to get involved. I realized that when I previously opened the issues there weren't any new good first issues (seems like there are some new ones now), so I took something old. I wanted to put up a small draft of it to see if it's actually relevant. Otherwise I'm happy to close and take another issue.

The changes are small and adds the location data object to the Agency model and specs. I also changed the equality check in the `Location` class. That said, I don't think there are any usages so I wonder if it's still relevant 😅 .

I didn't want to get too crazy with the specs, but I changed `be_truthy` to `eq(true)`. I think it's not a good practice: [relevant link if interested](https://gitlab.com/gitlab-org/gitlab/-/issues/230891).

Issue: https://github.com/EBWiki/EBWiki/issues/3389

In your PR did you:


app/models/location.rb
```diff
@@ -21,7 +21,12 @@ def to_s(letter_style: false)
 
   # rubocop:disable Style/RedundantSelf
   def ==(other)
-    other.class.eql?(self.class) && self.to_s.eql?(other.to_s)
+    other.class.eql?(self.class) &&
```
rlgreen91, line 24: Why did you make this change, especially since Rubocop has issues with the amount of branching conditions?

rlgreen91 (commented):
Hi @danielvdao, thanks for submitting your PR?  I just have a couple of questions, but once those are resolved we'll be good to go on merging your PR.

app/models/location.rb
danielvdao: Oh I did it because I thought it would be more correct to compare values as opposed to the stringified version of the class. But I'm happy to revert it back!

app/models/agency.rb
```diff
@@ -71,4 +71,15 @@ def slug_candidates
   def retrieve_state
     State.where(id: state_id).pluck(:name).join
   end
+
+  def location
+    Location.new(state: state.name, street_location: street_address, city: city, zipcode: zipcode)
```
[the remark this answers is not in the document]
danielvdao: Yes! Makes sense.

danielvdao:
@rlgreen91 I think this is ready for review. For some reason CodeClimate wasn't reporting back to CI. I just updated the branch, so hopefully it'll go through.

app/models/agency.rb
```diff
@@ -71,4 +71,20 @@ def slug_candidates
...
+  def location
+    @location ||= Location.new(
+      state: state.name,
+      street_location: street_address,
+      city: city,
+      zipcode: zipcode
+    )
+  end
+
+  def location=(location)
```
rlgreen91, line 84: Hmm, I'm not quite sure why you had to add this?

app/models/location.rb
rlgreen91: Yes, could you revert it back please?  In general, I think equivalent string values will be enough in this case.

rlgreen91 (commented):
Just the one question about an addition, and then we should be good to merge!

app/models/agency.rb
danielvdao: I was following the example from the issue [here](https://revs.runtime-revolution.com/value-objects-in-ruby-on-rails-9df64bc8db34) to create a setter method for `Location` value object. I think that's why the examples were non-memorized 🤔 , otherwise if it was memoized, then I think we would get the original `Location` object even if we updated an `Agency`'s address during runtime.

app/models/location.rb
danielvdao: Done!

app/models/agency.rb
rlgreen91: Oh, ok - in that case, it might be better not to memoize it.  Sorry, but do you mind removing it?  This is really is the last thing, haha.

danielvdao: Not a problem at all 🙂

danielvdao: Done! Thank you for responding so quickly.

rlgreen91 (approved):
Everything looks great!  Thanks so much for completing this - if you want to keep working on the feature this story is part of, then I'll probably write the issue tomorrow.  But if you just want to keep looking for some quick issues to tackle then we added a bunch around updating gems as well.

danielvdao:
@rlgreen91 I'd love to keep working on the feature if you write up the issue 🙂. I think there's another component to the issue as well (e.g. adding it to the `Organization` model so I'll get that one out this week as well), but I kept this one small so it'd be easier to review. I'll be happy to tackle other issues as I become more familiar with the project. Thank you again for reviewing!

rlgreen91:
Hey @danielvdao - just as a heads up, you don't have to worry about adding this to the `Organization` model just yet, we're going to put that on hold for now.

Issue #3947, opened 2021-02-21
Update rubocop gem


rlgreen91:
Update the rubocop gem to 0.75.  Note that you will need to update the corresponding channel listed in `.codeclimate-yml` to `rubocop-0-75`.

There's a pretty good chance that after the upgrade, running rubocop locally will reveal some fixes it wants to see.  If the changes are small and straightforward, feel free to go ahead and do them.  If you think something warrants a separate issue or more discussion, then just make a comment in your PR.

danielvdao:
Hey @rlgreen91 I think this PR should do it: https://github.com/EBWiki/EBWiki/pull/3979. That said, I think we're already on 0.93, I added some notes in the PR description.

danielvdao:
Follow-up to close this one all the way: https://github.com/EBWiki/EBWiki/pull/3981 🎉

PR #3979, opened 2021-03-11
Update rubocop to 0.93

danielvdao:
Issue: https://github.com/EBWiki/EBWiki/issues/3947

Brings rubocop up to 0.75. I could be wrong, but I think we're already running rubocop 0.93 production. [Gemfile.lock in main](https://github.com/EBWiki/EBWiki/blob/main/Gemfile.lock#L694).

When I bash into the container and run `bundle exec rubocop -v` I also see 0.93. That said, I upgraded the `.codeclimate.yml` so that we're running 0.75.

EDIT: After comments, looks like we're bringing it up to 0.93!


rlgreen91:
Thanks for tackling this @danielvdao!  Since that's the case, do you mind updating the Gemfile to list 0.93 as the version, and updating `codeclimate.yml` to use that version as well?

danielvdao:
@rlgreen91 done 🙂 . I could be wrong, but I think we also missed the rubocop-performance gem extension. I've also added that in.

mnyon-grandkru:
Hi @danielvdao , thanks for the work on this gem. Looking at the README for [rubocop-performance](https://github.com/rubocop/rubocop-performance), it looks like the recommended way to instantiate it would be to use `require:false` in the `Gemfile`, followed by one of the require techniques outlined in that README.  Would you mind requiring `rubocop-performance` with those techniques?  Thanks.

danielvdao:
Hi @mnyon-grandkru! Thanks for catching that. Total n00bie with the Gemfile so I learned something new today 🙂 . I went ahead and uncommented the `require` in `.rubocop.yml`. I cleaned up some of the rubocop exclusions we had along with removing one.

I turned off the excludes locally and cleaned up some of the specs we had issues with. Other ones would take a little longer so I avoided those for now. Maybe over time I can chip away at them.

danielvdao:
@rlgreen91 @mnyon-grandkru out of curiosity, why don't we have the latest rubocop? I see that we're pinned under the 1.11 major version. I don't see a potential conflict with any of the other gems we have, rubocop-performance is pinned to be under 2.0, but maybe I'm missing something?

mnyon-grandkru:
@rlgreen91 or @managedkaos might have more infromation on this. Updating to 1.11 doesn't disrupt the `bundle install` command, FWIW.

danielvdao:
I see. I think we should update rubocop all the way if possible as some of the changes that were made conflict with the later rubocop versions anyways (e.g. trailing comma in arguments). If we have agreement from @rlgreen91 and you / whomever else we may need it from, I'm happy to own that 🙂.

mnyon-grandkru:
@danielvdao I'm okay with upgrading to 1.11, but I'd put that in a separate issue/ PR. In that way, we can upgrade here as needed and move to 1.11 with everyone's input.

danielvdao:
@mnyon-grandkru Good point, I'm happy to do that separately.

mnyon-grandkru:
@danielvdao It looks like this has a failing spec in this [Github Action](https://github.com/EBWiki/EBWiki/runs/2096377212?check_suite_focus=true_).  Could you look into those and resolve them?

danielvdao:
@mnyon-grandkru that's a previous CI build, no? [This](https://github.com/EBWiki/EBWiki/pull/3979/checks?check_run_id=2096409089) is the most recent one I see 🤔

rlgreen91 (approved):
LGTM

PR #3981, opened 2021-03-12
Bring Rubocop up to latest version and add rubocop-performance

danielvdao:
### What
Brings rubocop and other gems up to the latest version. Enables rubocop-performance and enables rules as well. I think we are still missing one more gem though... which is `rubocop-rails` though maybe that could be saved for later.


danielvdao:
Rubocop is failing on some changes. Looks like I may have brought up other libraries by accident. Not sure if we should 100% roll out with it just yet, but the tests will pass so I think it's okay.

Otherwise, I can look at just upgrading rubocop.

danielvdao:
@rlgreen91 @mnyon-grandkru this is a DRAFT / WIP PR based off my last PR w/r/t rubocop. I have some thoughts and wondering if y'all agree, so I decided to open this PR more as a discussion to not block the last PR 🙂 .

1. Rubocop is behind, so I think we can bring it up all the way to the most recent version. Other gems were brought up as well 🤔 .
2. Enable all the rubocop rules and _retroactively_ exclude rubocop rules!
3. Rather than excluding directories specific, maybe I can try to add a exclude files and put files in them that match that pattern. This is so that we proactively enforce rubocop patterns and also allow us to go back and retroactively fix rubocop issues in the future. The end goal is that so future people (like myself) don't accidentally add rubocop issues and that we can clean things up slowly.

An example `.rubocop.yml` would look like:

```

inherit_from:
  - .rubocop_exclude.yml

inherit_mode:
  merge:
    - Exclude

require:
  - rubocop-rspec
  - rubocop-performance
  - rubocop-rails
```

And then in `.rubocop_exclude.yml` an example could be:

```
Layout/EmptyLinesAroundBlockBody:
  Exclude:
    - 'spec/**/*.rb'
```

rlgreen91:
Hey @danielvdao - I went ahead and merged your other PR.  Can you bring rubocop up to the current version in this PR?  And then when looking at how many corrections need to be made we can figure out if we should talk about your idea for the exclusions file in a separate issue/PR then.

danielvdao:
@rlgreen91 that makes sense. Sounds good, I'll see what rules get brought in / corrected 🙂.

danielvdao:
@rlgreen91 this is ready for review. I removed some files too because of the empty file rule and all of the specs passed. That said, I think there's another gem (`rubocop-rails`) which could help us too, but I wanted to separate that into another PR if you think it's worth it. I think it could help optimize and set good practices too.

rlgreen91:
Hey @danielvdao - that sounds good.  There are some merge conflicts though - can you resolve those and ping me again when it's ready?

danielvdao:
@rlgreen91 Thanks for taking a look. I resolved the conflicts, please let me know if I'm missing out on anything.

Gemfile
```diff
@@ -61,6 +61,9 @@ group :development, :test do
   gem 'mock_redis', '~> 0.26'
   gem 'pre-commit', '~> 0.39'
   gem 'rspec-rails', '~> 4.0', '>= 4.0.1'
+  gem 'rubocop', require: false
```
rlgreen91, line 64: @danielvdao as a heads up, we might have to move these back out because of an issue with requiring Rubocop in the Rakefile causing a `LoadError` during deployment.

rlgreen91 (approved):

PR #3993, opened 2021-03-21
Change case form cancel helper

rlgreen91:
This has nothing to do with a current issue and is really just something I was annoyed by.  Currently, the "back" button on the case form is technically a link, meaning that the text is purple once you have visited that URL.  Additionally, the button is also blue, rather than the gray used for negative actions elsewhere, even within the form.  This PR corrects that.  Whoever reviews this, if you approve of this change, then please create an issue assigned to me to make this change for all forms.

Also, the issue with Sprockets 4 and seg faults is still there, so I had to downgrade that as well.


mnyon-grandkru (approved):
LGTM

PR #4007, opened 2021-04-17
Removes rubocop from Rakefile

rlgreen91:
Removes rubocop from Rakefile, which in turns removes the rubocop rake task - this is an attempt to address a LoadError for rubocop during Heroku deploys.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,7 +2,6 @@
 
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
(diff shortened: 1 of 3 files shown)


mnyon-grandkru (approved):
LGTM. I’ve not used that rake task in awhile.

rlgreen91:
The failing spec does appear to be a flake, as it passed locally when using the same seed and a different one.  However, that spec tested that when an unauthenticated user attempts to go to the edit endpoint, the user is redirected to the login page - which is the default configuration for Devise.  There's not really a reason for us to test that because it's just testing that Devise works, so I removed those tests from this spec altogether.

PR #4034, opened 2021-05-30
Upgrades Rails to 5.1

rlgreen91:
Upgrades Rails to 5.1.  Fixes broken spec and bug for reverting a case

In your PR did you:


.github/workflows/ci.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:11 # try circleci/postgres:11-ram
+        image: postgres:12 # try circleci/postgres:11-ram
```
mnyon-grandkru, line 15: Good catch.

.rubocop.yml
```diff
@@ -325,3 +325,6 @@ Performance/StringInclude:
 
 Performance/Sum:
   Enabled: true
+
+Style/StringChars:
```
mnyon-grandkru, line 329: TIL about using `chars` insteaf.

Gemfile
```diff
@@ -6,7 +6,7 @@ source 'https://rubygems.org'
 git_source(:github) { |repo_name| "https://github.com/#{repo_name}.git" }
 
 gem 'dotenv-rails', '~> 2.7', groups: %i[development test production], require: 'dotenv/rails-now'
-gem 'rails', '~> 5.0.7.2'
+gem 'rails', '~> 5.1.7'
```
mnyon-grandkru, line 9: The meat & potatoes.

spec/requests/versions_spec.rb
```diff
@@ -17,7 +18,7 @@
 
     context 'reverts the version of the case' do
```
mnyon-grandkru, line 19: Can we add a spec to handle the failure case of the controller, when the `version.reify` returns nil or a `StandardError` is returned?

mnyon-grandkru (changes requested):
I've left a couple of comments that could use feedback, and this looks mostly nice!

spec/requests/versions_spec.rb
rlgreen91: Sure thing!

mnyon-grandkru: I was thinking of something like:
```
context 'when version.reify fails' do
      let(:version_mock) { double("version") }
      before do
        allow(PaperTrail::Version).to receive(find_by_id) { version_mock }
        post "/cases/#{this_case.id}/versions/#{version_id}/revert",
           params: {},
           headers: {
             "HTTP_REFERER": '/'
           }
      end

      it 'redirects to the previous page' do
        expect(response).to redirect_to("/cases/#{this_case.slug}")
      end

      it 'does not create a new version' do 
        # TODO: compare versions before & after method call
      end
    end
```

but that's not central to upgrading, so that can be worked on separately.

mnyon-grandkru (approved):
LGTM

PR #4059, opened 2021-10-02
Display characters left count for Blurb field in Case form

### LINKED ISSUE #4040: Display character counter for blurb section of case form
We currently use the blurb attribute on cases to store the text displayed on pins in map views of the cases.  Because of this, we limit blurbs to 500 characters.  In the new and edit case forms, add a UI element to warn users when they have 100 characters or less. Twitter's character limit element would be a good idea to emulate.

MohamedAsan:
In your PR did you:

**Issue:** Resolves #4040

**Description:**
Displays the number of characters left in the blurb field in new/edit case form
Turns to red when the characters left is lesser than 50

https://user-images.githubusercontent.com/29650842/135718518-494246d7-64e3-4404-ad63-7657bafb4e42.mov

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ logs:
 	docker logs --timestamps --follow ebwiki
 
 exec:
-	docker exec -it ebwiki bash
+	docker exec -e PGPASSWORD=ebwiki -it ebwiki bash
 
 stop:
 	@docker stop ebwiki > /dev/null 2>&1 || echo "ebwiki container is not running"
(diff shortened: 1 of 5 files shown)


rlgreen91 (approved):
Looks great @MohamedAsan, and thanks for adding the video as well!  Out of curiosity, did you try writing the Javascript code in a separate file?  We've had some issues in the past with the code not being Javascript not being run in that case, so I'm wondering if that was something you ran into.

MohamedAsan:
@rlgreen91 I didn't try adding JavaScript separately. Since the `<script>` tag was already present in the view, I thought it would be better to implement there itself

Issue #4060, opened 2021-10-02
Use Uglifier in harmony mode to enable ES6 syntax


rlgreen91:
Our production config is set to use Uglifier in the default mode for asset compilation.  In order to use ES6 syntax in our Javascript, we need to update the config to use Uglifier in harmony mode.

In `config/environments/production.rb`, update line 28 to be `config.assets.js_compressor = Uglifier.new(harmony: true)`

kinduff:
Happy to retake this one if #4064 is not going to be merged

rlgreen91:
@kinduff Sure thing

PR #4064, opened 2021-10-03
Issue 4060 Use Uglifier in harmony mode to enable ES6 syntax

a-developer-named-mark:
Made changes as outlined in https://github.com/EBWiki/EBWiki/issues/4060
Any more need please let me know.


rlgreen91:
Thanks for taking this on @a-developer-named-mark !  It seems that are failing specs in the CI run - `spec/requests/versions_spec`.  Can you see if you can fix those?  If you're unable to get the specs to fail locally, or unable to get them passing after an hour, just comment so here.

rlgreen91:
@a-developer-named-mark are you able to just skip the failing specs then?

PR #4066, opened 2021-10-06
Correctly Select Subject Race/Gender Values on Edit

### LINKED ISSUE #3963: Gender and Race dropdowns default to first values in edit form
When editing a case, the gender and race/ethnicity dropdown fields show the first value as the selected item.  Instead, these fields should show the actual value saved to the case as the selected item.

To reproduce:

- Create a case with a subject and choose "female" as the gender selection and "Asian" as the race/ethnicity selection.
- Navigate to the edit view of the case
- View the dropdowns for the gender and race/ethnicity selections.  They will default to the first option but should instead show the options you selected above.

yez:
[Fixes #3963]

When editing a case, use the subject's gender_id and ethnicity_id values
to autoselect the drop down values.

Gif of changing values, navigating to the home page and then seeing them autopopulate in the edit form. 

dropdown

In your PR did you:

### DIFF
--- a/spec/models/agency_spec.rb
+++ b/spec/models/agency_spec.rb
@@ -3,45 +3,42 @@
 require 'rails_helper'
 
 RSpec.describe Agency, type: :model do
-  before(:each) do
-    @texas = FactoryBot.create(:state_texas)
+  let(:texas) { FactoryBot.create(:state_texas) }
+  let(:ohio) { FactoryBot.create(:state_ohio) }
+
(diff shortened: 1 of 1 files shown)


yez:
Looks like the rspec build failed but I am not sure why those tests would be impacted by this change. I don't think I'm able to retry the specs (in the case where they're flakey). I'm not able to consistently replicate failures locally. I see different failures on some runs.

Issue #4072, opened 2021-10-09
Create a rollout value object


rlgreen91:
Create `app/models/rollout.rb`. This will be a value object, not an ActiveRecord model - a lot like the `DateRange` object.

The `rollout` object should:
- receive a path for a YAML file when initialized
- should use the data in the YAML file to populate `name`, `description`, `creator`, and `date_added` attributes
- respond to `to_s` with the name attribute converted to a string
- should use the same definition of the comparison method as the `DateRange` object.

olikristen:
Hi ! I would take it. Can you assign it to me? :)

rlgreen91:
Not at the moment, but if you check out this issue you can see what fields we're planning to define in the file: Issue #4062

olikristen:
Awesome! Please check my PR for the feature and let me know if there is anything to adjust.

rlgreen91:
@olikristen just as a heads up, I updated this ticket.  Now that we have a proper `ValueObject` class you can have the rollout object inherit from that and remove the `==` method definition.

PR #4074, opened 2021-10-10
[Issue 4073] Add value object Rollout

olikristen:
This PR adds the value object rollout which reads a YAML file. This PR is related to the issue #4072 .


rlgreen91:
Hi @olikristen, it looks like you have a few Rubocop errors - can you address those?  You can see the list of them in the "Checks" tab of the PR.  You can just disable the check around the `==` method like we did in the `DateRange` object.

PR #4078, opened 2021-10-13
Value Object class

### LINKED ISSUE #4073: Create a value object class
Create `app/models/value_object.rb`, which contains the `ValueObject` class.  This class should **not** be an ActiveRecord model.  The class should contain a `to_s` method that raises a `NotImplementedError`.  It should also contain the following method definition:
```
 # rubocop:disable Style/RedundantSelf
  def ==(other)
    self.class == other.class && self.to_s == other.to_s
  end
  # rubocop:enable Style/RedundantSelf
```

Update the `DateRange` and `Location` objects to inherit from `ValueObject`.  You can remove the `==` method from each of those classes.

kinduff:
Closes #4073 

Implements the `ValueObject` class to be parent of `Location` and `DateRange`. This is going to be useful in the future in the case we want to share methods.

Wanted to use a Shared Example for the equality `#==` method but since each subclass implements their own to string `#to_s` method, it caused more overhead than leaving the tests as is.

Had to disbale the `Lint/MissingSuper` cop since it doesn't make sense for now.

In your PR did you:


kinduff:
Specs sometimes randomly fail. My suggestion is to start saving the seeds and failed examples so we can reproduce and fix in the future.

PR #4098, opened 2021-10-27
Require new specs for new .rb files

eric-kovshovik:
This feature will fail rspec/pipeline if there are no new specs for a newly added .rb file(helper/service/etc).  
Also I'd love to rename `requests` folder to `controllers` for consistence in testing. (so the same script can be used with controllers). What do you guys thinks about it?

In your PR did you:


mnyon-grandkru:
@rlgreen91 @eric-kovshovik Does this pull request and issue still spark joy, or should we let this go?

